PR #1, opened 2016-10-10
Initial Build Command

matt-gadd:
**Type:** feature
This is the initial build command using webpack. At the moment it supports:
production build: `dojo build` runs the build and generates a distribution in `dist` 
dev build: `dojo build -w` runs a watcher and livereloads in the browser automatically

We will likely want to expand the webpack config itself as we go along, to deal with CSS modules and other items, this was just the basic amount to get todo-mvc compiling originally.

* [x] There is a related issue https://github.com/dojo/cli/issues/15


tests/support/MockModule.ts
```diff
@@ -0,0 +1,81 @@
+import * as mockery from 'mockery';
```
agubler: Candidate to be moved to a sharable module across all our packages 👍

PR #8, opened 2016-10-20
Support sourcemaps for dojo 2 dependencies

matt-gadd:
**Type:** feature

**Description:**

Adds support of sourcemaps back to original TS (rather than just the unminified JS) for our dojo 2 dependencies. Unfortunately the official webpack source-map-loader looks dead as a project, with an outstanding PR that fixes up the sourcemap source paths, so I've had to use a fork.


dylans:
Interesting that webpack has sat on https://github.com/webpack/source-map-loader/pull/9 for over a year...

Issue #9, opened 2016-10-24
External CSS


matt-gadd:
**Enhancement** 

The current webpack config internalizes the css into the js bundle. This although cuts down on network requests, causes a flicker of unstyled content on page load, as well as increasing the size of the bundle.

PR #10, opened 2016-10-24
Extract CSS to separate file rather than loading from bundle

matt-gadd:
**Type:** feature

**Description:**

Fixes https://github.com/dojo/cli-build/issues/9

Styles are now extracted to a separate `main.css` file, and source maps are generated. The html link tag for the stylesheet will be automatically injected by the HTML plugin.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
   "dependencies": {
     "copy-webpack-plugin": "^3.0.1",
     "css-loader": "^0.25.0",
+    "extract-text-webpack-plugin": "^1.0.1",
     "file-loader": "^0.9.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -44,7 +45,7 @@
     "stylus": "^0.54.5",
     "stylus-loader": "^2.3.1",
     "ts-loader": "^0.8.2",
-    "umd-compat-webpack-plugin": "1.0.2",
+    "umd-compat-webpack-plugin": "1.0.4",
```
matt-gadd: this needs a later version of the umd-compat plugin, as it was seemingly injecting js into entries that weren't js.

PR #15, opened 2016-11-07
CSS Modules

bitpshr:
**Type:** Feature

**Description:**

This PR adds optional CSS module compilation to the build command. After some discussion with @matt-gadd, @kitsonk, and @agubler, this seemed to make more sense than a separate CSS module CLI command. Note that an intermediary `dojo css typings` command has been developed at https://github.com/bitpshr/cli-css-typings that would theoretically be used during development in order to leverage intellisense.

A user may specify a `cssModules` config in the webpack configuration file:

```js
// ...
cssModules: {
    root: path.join(basePath, 'src/styles/structural/*.css'),
    cssOut: path.join(basePath, 'src/styles/structural/_css/'),
    tsOut: path.join(basePath, 'src/styles/structural/_modules/')
}
// ...
```

Resolves dojo/widgets#67, dojo/cli#61, dojo/widgets#66, dojo/meta#28

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,12 +34,18 @@
     "yargs": "^5.0.0"
   },
   "dependencies": {
+    "async": "^2.1.2",
+    "chalk": "^1.1.3",
     "copy-webpack-plugin": "^3.0.1",
     "css-loader": "^0.25.0",
     "extract-text-webpack-plugin": "^1.0.1",
     "file-loader": "^0.9.0",
+    "globby": "^6.1.0",
     "html-loader": "^0.4.4",
     "html-webpack-plugin": "^2.22.0",
+    "mkdirp": "^0.5.1",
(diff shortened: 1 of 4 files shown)


matt-gadd:
I think the idea of making css-modules part of this (and having a separate author time typings command for intellisense) is sound, but I think we should do the build part via a loader or plugin for webpack due to the following:

`dojo-build-webpack` is a specific webpack implementation of the build pipeline. Users eventually will be able to eject the config (https://github.com/dojo/cli/issues/44)  and get the same results as `dojo build` via just webpack itself. This will no longer be the case if we are now implementing specific things in the actual command itself.

If this was implemented as a webpack plugin/loader this would work correctly with the `watch` capabilities of `webpack-dev-server` also, which this currently doesn't.

PR #32, opened 2016-12-13
Implement app-level config loading.

### LINKED ISSUE #29: Allow app-level config for basic overrides/input.
While updates to the webpack configuration itself should and will be handled by the `eject` command (dojo/cli#44), certain options justify an application-level config. For example, the user should be able to set the application's default locale and any supported locales. As such, the build process should check for the existence of something like `path.join(process.cwd(), 'dojo.config.js')` and read any values it needs to feed into the configuration.

mwistrand:
**Type:** feature

**Description:** Provides the ability to load an application-level configuration object that will eventually be used to pass app-specific data to the build. If there is no config file in the specified location, then a default config is returned. At present the default config is an empty object, and it is not being loaded by the webpack config as there is nothing that currently needs it. Upcoming features like #30 will use the the app config to determine the final build.

Resolves #29

### DIFF
--- a/tests/unit/loadDojoConfig.ts
+++ b/tests/unit/loadDojoConfig.ts
@@ -1,0 +1,18 @@
+import { describe, it } from 'intern!bdd';
+import * as assert from 'intern/chai!assert';
+import loadDojoConfig from '../../src/loadDojoConfig';
+
+describe('loadDojoConfig', () => {
+	it('should load the specified module if it exists', () => {
+		const config = loadDojoConfig('tests/support/mocks/dojo.config.js');
+		assert.deepEqual(config, {
+			foo: 'foo',
+			bar: 'bar'
+		});
+	});
+
+	it('should load an empty object if no config exists', () => {
+		const config = loadDojoConfig('path/to/nonexistent/dojo.config.js');
+		assert.deepEqual(config, {}, 'A default config is returned.');
+	});
+});
--- a/package.json
+++ b/package.json
@@ -17,14 +17,16 @@
     "test": "grunt test"
   },
   "devDependencies": {
+    "@types/node": "0.0.3",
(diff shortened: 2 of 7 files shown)


tests/unit/loadDojoConfig.ts
```diff
@@ -0,0 +1,18 @@
+import { describe, it } from 'intern!bdd';
+import * as assert from 'intern/chai!assert';
+import loadDojoConfig from '../../src/loadDojoConfig';
+
+describe('loadDojoConfig', () => {
+	it('should load the specified module if it exists', () => {
+		const config = loadDojoConfig('tests/support/mocks/dojo.config.js');
```
mwistrand: In that case, would `app.config.js` be preferable?

matt-gadd: will this only contain build configuration? i'm keen to avoid N amount of configs in the root of a project, I was hoping we could have a `.dojorc` that would potentially contain configuration for cli commands in general.

matt-gadd: and if that was the case cli-build wouldn't have to worry about this at all, it would just be received in the command.

mwistrand: Yes. The idea is that there should only be one config, and `webpack.config.js` would pull in just this one file. Honestly, passing a filename to the function instead of hard-coding a path serves in part to facilitate testing (so that a config does not need to be generated at the root of the cwd and then removed after testing).

matt-gadd: I've raised an issue here https://github.com/dojo/cli/issues/88, to solve this generically in cli - let me know if that makes sense.

mwistrand:
Closing in favor of `dojo/cli#88`.

PR #33, opened 2016-12-14
Add dojo-core/load, inject modules, and dojo-i18n plugins.

### LINKED ISSUE #30: Implement i18n into build process
@rishson commented on [Tue Oct 04 2016](https://github.com/dojo/cli/issues/60)

Refs dojo/meta#55

---

@mwistrand commented on [Sat Oct 29 2016](https://github.com/dojo/cli/issues/60#issuecomment-257105335)

Since `dojo-i18n` allows multiple locales to be used simultaneously and for locale data to be loaded dynamically, the main obstacle to incorporating `dojo-i18n` into the build process is ensuring that the CLDR data and locale messages are loaded only once.

The CLDR data are loaded via `dojo-core/request` and then cached locally, and the locale messages are loaded via `dojo-core/load` and then cached locally. Since the locale messages are proper modules, theoretically they should pose no problem since the build will have already loaded them. The CLDR data, however, are JSON files. As such, pointing to the appropriate cache in both development and production is a more difficult problem. One solution is to write an additional module that converts the CLDR JSON files pulled from unicode.org into UMD modules, so that they can be imported by `dojo-core/load` and cached by the underlying loader.

With regard to webpack, @matt-gadd, @agubler, @tomdye, and I had discussed using [`globalize-webpack-plugin`](https://github.com/rxaviers/globalize-webpack-plugin), but found that doing so would require changes to the `dojo-i18n` message structure and would mean loading CLDR and message data twice in development.

Whichever solution we settle on, the following are still required:
1. A default locale.
2. An array of locales that should be included in the final build. Any default locale that has been specified will be assumed and not required in this list.
3. An array of paths to bundle directories. Any files found under `${path}/{locale}/` (where `{locale}` is one of the locales listed in no. 2 above) will be included in the build.
4. All CLDR data for the locales listed in no. 2 above are included in the build.
5. It should still be possible to dynamically load any other CLDR data or messages for locales not explicitly included in the list of supported locales.

mwistrand:
**Type:** feature

**Description:**

Resolves #30 

Blocked by dojo/i18n#33 and dojo/core#249.

Implements the following webpack plugins:

1. `CoreLoadPlugin`, which replaces `dojo-core/load` with `dojo-core/load-webpack`, injecting a map of module paths to webpack mids into the former, and instructing webpack to ignore contextual requires.
2. `InjectModulesPlugin`, which injects arbitrary modules into the build, specifically into chunks where a specified module is found.
3. `I18nPlugin`, which includes message bundles and CLDR data for the specified locale(s) into the build.

Adds the following `yarg` options:

1. `locale`: the default locale for the application.
2. `supportedLocales`: an optional array of locales beyond the default to include in the build.
3. `messageBundles`: an optional array of message bundle paths. All bundles for the `supportedLocales` will be included in the build.

What's missing:

- A means of lazy-loaded additional locales not included in the main chunk.
- A call to `dojo-i18n/i18n#switchLocale` from within the build. Since `switchLocale` returns a promise, there needs to be a way to call it with one of the supported locales and ensure the application does not start until it resolves. For the time being, `switchLocale` needs to be called from within the application files.


src/plugins/CoreLoadPlugin.ts
```diff
@@ -0,0 +1,121 @@
...
+	 * @param compiler
+	 * The compiler instance.
+	 */
+	apply(compiler: any) {
+		const idMap = Object.create(null) as ModuleIdMap;
+		const basePath = compiler.options.resolve.root[0];
+		const bundleLoader = /bundle.*\!/;
+		const issuers: string[] = [];
+
+		compiler.apply(new NormalModuleReplacementPlugin(/dojo-core\/load\.js/, 'dojo-core/load-webpack'));
```
matt-gadd: is this now `dojo-core/load/webpack`?

mwistrand: Yes; and I'm updating the i18n `load-webpack.ts` to the same.

mwistrand:
With the alpha.4 release of [`dojo-i18n`](https://github.com/dojo/i18n), this PR is only blocked until the next release of `dojo-core` that will include the [`core/load/webpack`](https://github.com/dojo/core/commit/55153b350a34693d9e3eecf8b5b6b1ac9a25d748) required for this build to pass.

mwistrand:
Now that `async`/`await` can be down-emitted to ES5, we can automatically switch the locale by injecting something like the following into the top-level module:

```
(async function () {
    switchLocale(defaultLocale);
})();
```

matt-gadd:
@mwistrand I think @agubler owe's you a drink - because I think he might have broken parts of this now with the move to the `@dojo` namespace?

mwistrand:
@matt-gadd Since it seems #40 will land in the very near future, I'll wait and then update when it does.

PR #37, opened 2017-01-09
Build test files

maier49:
**Type:** bug / feature

**Description:**

Adds a flag to build test files, and fixes issues with unit tests caused by modules being cached between tests. Related to: https://github.com/dojo/cli/issues/13

### DIFF
--- a/src/webpack.config.js
+++ b/src/webpack.config.js
@@ -83,11 +83,23 @@ module.exports = function (args) {
     };
 
     if (args.withTests) {
-        plugins.push(new CopyWebpackPlugin([
-            {context: 'tests', from: '**/*', ignore: '*.ts', to: '../_build/tests' },
-        ]));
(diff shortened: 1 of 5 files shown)


src/webpack.config.js
```diff
@@ -32,44 +32,101 @@ module.exports = function (args) {
...
+                { test: /\.css$/, loader: 'style-loader!css-loader?modules' },
+            ]
+        },
+        plugins: plugins,
+        output: {
+            path: path.resolve('./dist'),
+            filename: '[name].js'
+        }
+    }];
+	if (args.withTests) {
```
matt-gadd: would it not be better to mix the below over the config rather then duplicating?

maier49: I added in `webpack-config` and used it to extract the common configuration.

maier49: This has been updated to use a single config.

PR #44, opened 2017-01-16
Update to scoped packages

agubler:
**Type:** bug

**Description:**

update references to scoped packages.

### DIFF
--- a/src/plugins/CoreLoadPlugin.ts
+++ b/src/plugins/CoreLoadPlugin.ts
@@ -70,7 +70,7 @@ export default class DojoLoadPlugin {
 		const bundleLoader = /bundle.*\!/;
 		const issuers: string[] = [];
 
-		compiler.apply(new NormalModuleReplacementPlugin(/@dojo/core\/load\.js/, resolveMid('@dojo/core/load/webpack')));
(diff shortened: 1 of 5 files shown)


src/plugins/CoreLoadPlugin.ts
```diff
@@ -70,7 +70,7 @@ export default class DojoLoadPlugin {
 		const bundleLoader = /bundle.*\!/;
 		const issuers: string[] = [];
 
-		compiler.apply(new NormalModuleReplacementPlugin(/dojo-core\/load\.js/, resolveMid('dojo-core/load/webpack')));
+		compiler.apply(new NormalModuleReplacementPlugin(/@dojo/core\/load\.js/, resolveMid('@dojo/core/load/webpack')));
```
mwistrand: This needs an escape after `dojo`: `/@dojo/core` => `/@dojo\/core`.

agubler: Course, thanks.

agubler: @mwistrand does the `@` need escaping too?

mwistrand: `@` is not a control/special character, so no, it should not need to be escaped.

agubler: i realised that after I said it, but thank you again.

mwistrand:
This issue (and code from this PR) merged with #45.

PR #47, opened 2017-01-17
Keep global require's toUrl and toAbsMid methods intact

maier49:
**Type:** bug

**Description:**

Currently, if code is calling `require.toUrl`, that will trigger the `expression require` condition and that module will have `require = function() { return '${path}'; }` prepended to it. However, this function obviously doesn't have a `toUrl` property so the call will fail. Checking for usages of `require.toUrl` and not replacing require at all is problematic because webpack will still replace it with `__webpack_require__` which doesn't have a `toUrl` function either.

It seems like there should probably be a better way to handle this. But this does work for todoMVC at least.

### DIFF
--- a/src/plugins/CoreLoadPlugin.ts
+++ b/src/plugins/CoreLoadPlugin.ts
@@ -83,7 +83,17 @@ export default class DojoLoadPlugin {
 			compilation.moduleTemplate.plugin('module', (source: any, module: any) => {
 				if (module.meta && module.meta.isPotentialLoad) {
 					const path = stripPath(basePath, module.userRequest);
-					const require = `var require = function () { return '${path}'; };`;
+					const require = `var require = (function () {
+						var globalScope = typeof window === 'undefined' ? global : window;
+						var toUrl = globalScope && globalScope.require && globalScope.require.toUrl
+						&& globalScope.require.toUrl.bind(globalScope.require);
+						var toAbsMid = globalScope && globalScope.require && globalScope.require.toAbsMid
+						&& globalScope.require.toAbsMid.bind(globalScope.require);
+						var newRequire = function () { return '${path}'; }; 
+						newRequire.toUrl = toUrl;
+						newRequire.toAbsMid = toAbsMid;
+						return newRequire;
+					})();`;
 					return new ConcatSource(require, '\n', source);
 				}
 				const load = idMap['@dojo/core/load'] || { id: null };
--- a/src/plugins/CoreLoadPlugin.ts
+++ b/src/plugins/CoreLoadPlugin.ts
@@ -84,11 +84,25 @@ export default class DojoLoadPlugin {
 				if (module.meta && module.meta.isPotentialLoad) {
 					const path = stripPath(basePath, module.userRequest);
 					const require = `var require = (function () {
-						var globalScope = typeof window === 'undefined' ? global : window;
-						var toUrl = globalScope && globalScope.require && globalScope.require.toUrl
-						&& globalScope.require.toUrl.bind(globalScope.require);
-						var toAbsMid = globalScope && globalScope.require && globalScope.require.toAbsMid
-						&& globalScope.require.toAbsMid.bind(globalScope.require);
+						var globalObject = (function () {
+							if (typeof window !== 'undefined') {
+								// Browsers
+								return window;
+							}
+							else if (typeof global !== 'undefined') {
+								// Node
+								return global;
+							}
+							else if (typeof self !== 'undefined') {
+								// Web workers
+								return self;
+							}
+							return {};
+						})();
+						var toUrl = globalObject && globalObject.require && globalObject.require.toUrl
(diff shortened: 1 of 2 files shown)


src/plugins/CoreLoadPlugin.ts
```diff
@@ -83,7 +83,17 @@ export default class DojoLoadPlugin {
 			compilation.moduleTemplate.plugin('module', (source: any, module: any) => {
 				if (module.meta && module.meta.isPotentialLoad) {
 					const path = stripPath(basePath, module.userRequest);
-					const require = `var require = function () { return '${path}'; };`;
+					const require = `var require = (function () {
+						var globalScope = typeof window === 'undefined' ? global : window;
```
dylans: We do seem to have several ways in Dojo 2 of getting a reference to the global, whether it is https://github.com/dojo/shim/blob/79882ec5c4fa78eb91171e7ff42ace275aa68562/src/support/global.ts or https://github.com/dojo/loader/blob/master/src/loader.ts#L8 . Maybe we want to normalize this?

tests/unit/plugins/CoreLoadPlugin.ts
```diff
@@ -62,7 +62,17 @@ describe('core-load', () => {
 		})[0];
 
 		assert.instanceOf(source, ConcatSource, 'A new `ConcatSource` is created.');
-		assert.strictEqual(source.source(), `var require = function () { return '/root/path/src/module'; };\n`,
+		assert.strictEqual(source.source(), `var require = (function () {
+						var globalScope = typeof window === 'undefined' ? global : window;
```
dylans: Same comment/question here about reference to global.

src/plugins/CoreLoadPlugin.ts
maier49: Normalizing the usage definitely seems like a good idea. I've updated the usage here to match the loader example, since that was more or less equivalent to what this was already doing but more thorough.

mwistrand:
Where is `globalObject.require` being set? If an application is being built with webpack and therefore using its internal loader, there would be no `require` in the global namespace.

maier49:
This came up within the context of building our tests. When we run the tests with Intern it's still going to use its own loader so there will be a `require` in the global namespace. 
If there isn't one this effectively reverts back to its original behavior.
-------- Original message --------From: Matt Wistrand <notifications@github.com> Date: 1/17/17  5:45 PM  (GMT-05:00) To: dojo/cli-build <cli-build@noreply.github.com> Cc: Bradley Maier <brdlmaier49@gmail.com>, Author <author@noreply.github.com> Subject: Re: [dojo/cli-build] Keep global require's toUrl and toAbsMid methods
  intact (#47) 
Where is globalObject.require being set? If an application is being built with webpack and therefore using its internal loader, there would be no require in the global namespace.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub, or mute the thread.


  
  




{"api_version":"1.0","publisher":{"api_key":"05dde50f1d1a384dd78767c55493e4bb","name":"GitHub"},"entity":{"external_key":"github/dojo/cli-build","title":"dojo/cli-build","subtitle":"GitHub repository","main_image_url":"https://cloud.githubusercontent.com/assets/143418/17495839/a5054eac-5d88-11e6-95fc-7290892c7bb5.png","avatar_image_url":"https://cloud.githubusercontent.com/assets/143418/15842166/7c72db34-2c0b-11e6-9aed-b52498112777.png","action":{"name":"Open in GitHub","url":"https://github.com/dojo/cli-build"}},"updates":{"snippets":[{"icon":"PERSON","message":"@mwistrand in #47: Where is `globalObject.require` being set? If an application is being built with webpack and therefore using its internal loader, there would be no `require` in the global namespace."}],"action":{"name":"View Pull Request","url":"https://github.com/dojo/cli-build/pull/47#issuecomment-273325084"}}}

matt-gadd:
I don't think we should land this in its current form for the following reasons:

* The current require code in `cli-build` gives context require support to webpack without an AMD loader giving parity in functionality. For `toUrl` and `toAbsMid` in this, that would not be the case which would be confusing in my opinion.

* This boilerplate would be added to every point a contextual require is made (regardless of whether `.toUrl` or `.toAbsMid` is used) which seems like a overhead in bundle size.

* You can actually use `require.toUrl` with a contextual require in AMD - which this doesn't support.

When we move to the version of intern that is loaderless, we won't have an AMD loader anyway, so it seems like in the mean time we might be better just changing the way we access an html page in todo-mvc to not use `require.toUrl`?

maier49:
@matt-gadd I agree with that reasoning. 

It does bring up one other concern though. This code is currently triggering on usage of `require.toUrl`. If it actually was a contextual require it will still be replaced with a function without a `toUrl` method, meaning that using `require.toUrl` with a contextual require is not supported by the current version either. Unless I'm missing something in how that would work.

PR #20, opened 2016-11-11
Enable Stylus support

bitpshr:
**Type:** feature

The following has been addressed in the PR:

* [x] There is a related issue
* [x] All code matches the [style guide](https://github.com/dojo/meta/blob/master/STYLE.md)
* [ ] Unit or Functional tests are included in the PR

**Description:**

This PR modifies the webpack config so CSS modules can be written in Stylus and depends on dojo/cli-css-typings#5. Note that to differentiate between regular `.styl` files and CSS module `.styl` files, by convention, CSS module names should end in `.module.styl`, e.g. `example.module.styl`.

Resolves #19

PR #48, opened 2017-01-18
Feature - Webpack build changes to support css modules

Tomdye:
**Type:** feature

**Description:**

- Adds ability to generate css-module output for projects using the CLI.
- IDE changed spaces to tabs

Resolves: https://github.com/dojo/widgets/issues/224
Supersedes: https://github.com/dojo/cli-build/pull/20 

Note: Typed CSS modules are outside the scope of this initial delivery


src/webpack.config.js
```diff
@@ -48,76 +56,94 @@ module.exports = function (args) {
...
+			libraryTarget: 'umd',
+			path: path.resolve('./dist'),
+			filename: '[name].js'
+		},
+		devtool: 'source-map',
+		resolve: {
+			root: [ basePath ],
+			extensions: ['', '.ts', '.tsx', '.js', '.css.json'],
+			alias: {
+				rxjs: '@reactivex/rxjs/dist/amd'
```
dylans: Do we need this any more?

src/webpack.config.js
```diff
@@ -48,76 +56,94 @@ module.exports = function (args) {
...
+		],
+		output: {
+			libraryTarget: 'umd',
+			path: path.resolve('./dist'),
+			filename: '[name].js'
+		},
+		devtool: 'source-map',
+		resolve: {
+			root: [ basePath ],
+			extensions: ['', '.ts', '.tsx', '.js', '.css.json'],
```
dylans: Probably don't need tsx, but maybe better to include it?

src/webpack.config.js
Tomdye: Not sure, but I don't think removing it would be in the scope of this PR

src/webpack.config.js
Tomdye: not sure why it's there, but again probably not a change that's in the scope of this PR

PR #61, opened 2017-02-02
Adding options to build components

rorticus:
**Type:** feature

**Description:**

Adding additional options to generate custom element packages and imports. 

Usage:

```bash
$ dojo build --element=src/widget/createMyWidgetElementDescriptor
```

This will bundle all of the dependencies from `createMyWidgetElementDescriptor` (usually your widget) and create the files,

* `dist/my-widget/my-widget.js` - JS for your widget
* `dist/my-widget/widget-core.js` - JS for widget core
* `dist/my-widget/my-widget.css` - CSS for your widget
* `dist/my-widget/my-widget.html` - HTML import containing all your widget dependencies

There is an optional parameter to name the output files, `--elementPrefix`, that you can specify if your input file does not follow the convention `createXYZElement`.

To use in an HTML file,

```html
<head>
    <script src="https://rawgit.com/webcomponents/custom-elements/master/src/native-shim.js"></script>
    <link rel="import" href="/path/to/myWidget.html">
</head>
<body>
    <my-widget></my-widget>
</body>
```

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,5 +1,7 @@
 module.exports = function (grunt) {
-	require('grunt-dojo2').initConfig(grunt, {});
+    require('grunt-dojo2').initConfig(grunt, {
+        staticDefinitionFiles: ['**/*.d.ts', '**/*.html']
+    });
 	grunt.registerTask('ci', [
 		'intern:node'
 	]);
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,8 +1,8 @@
-module.exports = function (grunt) {
+module.exports = function(grunt) {
     require('grunt-dojo2').initConfig(grunt, {
         staticDefinitionFiles: ['**/*.d.ts', '**/*.html']
     });
-	grunt.registerTask('ci', [
-		'intern:node'
-	]);
+    grunt.registerTask('ci', [
+        'intern:node'
+    ]);
 };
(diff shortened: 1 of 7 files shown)


Gruntfile.js
```diff
@@ -1,5 +1,7 @@
 module.exports = function (grunt) {
-	require('grunt-dojo2').initConfig(grunt, {});
+    require('grunt-dojo2').initConfig(grunt, {
```
agubler: spaces vs tabs?

agubler:
General question do you think it should output to a directory in `dist` using the name of the component?

agubler:
Silly codecov, do you think we should exclude template files?

rorticus:
> Silly codecov, do you think we should exclude template files?

Yes!

rorticus:
> General question do you think it should output to a directory in dist using the name of the component?

I'm not a fan of the `elementPrefix` option. Ideally it would be able to read the `tagName` from the element descriptor factory, but I'm not sure how to do that without opening the file and reading it with a regex (would that be ok?!).  If we could do that, we could just do something like,

```bash
$ dojo build --element=/path/to/factory.ts
```

And the magic would happen.

Also, I am totally ok with outputting to a directory based on the element name!

agubler:
@rorticus feels like we can work on the magic, but it would be nicer to just point to the element factory. 👍 for building into a directory for the widget.

rorticus:
Rebased onto master, outputting files to `dist/xyz` instead of `dist`, and added `elementPrefix` auto detection!

Issue #53, opened 2017-01-25
Type the WebPack config


matt-gadd:
Now we're moving the WebPack config over to TypeScript, we should make an effort to properly type it.

Issue #68, opened 2017-02-20
Invocation of `Compilation#buildModule()` is wrong for Webpack 2


bryanforbes:
The API for `Compilation#buildModule()` changed in Webpack 2 and `InjectModulesPlugin` was not updated to accommodate this change. Currently, the following is done:

```ts
compilation.buildModule(module, (error?: Error) => {
```

But it should be called like so:

```ts
compilation.buildModule(module, false, null, null, (error?: Error) => {
```

I discovered this while writing typings for Webpack for #53.

bryanforbes:
I will submit a fix for this issue as part of the pull request to fix #53.

mwistrand:
This was resolved by #69.

PR #69, opened 2017-02-20
Webpack typings

bryanforbes:
**Type:** enhancement

**Description:**

This adds typings for Webpack and related libraries while also updating the code to conform to the typings and fixing any bugs in API usage.

Resolves: #53, #68


src/webpack.config.ts
```diff
@@ -111,33 +116,35 @@ module.exports = function (args: any) {
...
 			}),
 			new CoreLoadPlugin(),
-			...includeWhen(args.element, (args: any) => {
-				return [ new webpack.optimize.CommonsChunkPlugin('widget-core', 'widget-core.js') ];
+			...includeWhen(args.element, () => {
+				return [ new webpack.optimize.CommonsChunkPlugin({ name: 'widget-core', filename: 'widget-core.js' }) ];
 			}),
-			new webpack.optimize.UglifyJsPlugin({ sourceMap: true, compress: { warnings: false }, exclude: /tests[/]/ }),
-			includeWhen(args.element, (args: any) => {
+			/*new webpack.optimize.UglifyJsPlugin({ sourceMap: true, compress: { warnings: false }, exclude: /tests[/]/ }),*/
```
matt-gadd: I presume this isn't meant to be commented out :)

src/loaders/css-module-dts-loader/loader.ts
```diff
@@ -73,7 +74,7 @@ function traverseNode(node: Node, filePaths: string[] = []): string[] {
 	return filePaths;
 }
 
-export default function (this: Webpack, content: string, sourceMap?: string) {
+export default function (this: webpack.LoaderContext, content: string, sourceMap?: string) {
```
tomdye: Worth deleting my crude Webpack typing if it's no longer required

src/webpack.d.ts
```diff
@@ -0,0 +1,738 @@
+declare module 'tapable' {
+	class Tapable {
```
tomdye: Did you type out all this?

src/loaders/css-module-decorator-loader/loader.ts
```diff
@@ -1,6 +1,8 @@
+import webpack = require('webpack');
```
tomdye: Can this not use the import from syntax?

src/webpack.d.ts
bryanforbes: Yes. I took a look at the `webpack` and `tapable` typings in `@types` and they were terrible.

tomdye: Just wasn't sure why we'd want / need to maintain our own.

src/loaders/css-module-decorator-loader/loader.ts
bryanforbes: I believe we can use the import/from syntax, but I've run into issues where importing from a module typed with `export =` triggers a compile error (and, of course, I can't find any examples right now). Because of that, I tried to standardize on using import/require for modules that export a function and import/from for ES6-like modules.

PR #72, opened 2017-02-23
CSS-Module Dts preloader for TS and CSS files

### LINKED ISSUE #66: Generate css-module d.ts files as part of build command.
`css-module-typings` should be created as part of our webpack build process. Furthermore they should be included in the watch process such that they are generated as they are edited.
An attempt was made to use `typed-css-module-loader` but this was not sufficient. There are outstanding PR's and issues on this loader which may provide a solution but it's likely that we may need to create a `typescript` pre-loader which plucks out the `.css` imports and creates the appropriate typings.

tomdye:
**Type:** feature

**Description:**
Adds CSS and TS preloader to generate the required css-module dts files.

Resolves #66


src/loaders/css-module-dts-loader/loader.ts
```diff
@@ -0,0 +1,125 @@
...
+		else {
+			return Promise.resolve();
+		}
+	}
+	else {
+		return Promise.resolve();
+	}
+}
+
+function checkNode(node: Node, instance?: TSLoaderInstances): Promise<any> {
```
matt-gadd: feels like this and `checkNodeForCSS` should all be sync and the only thing async would be `generateDTSFile`?

matt-gadd: let me clarify, I think it would be easier to understand if you performed all the sync traversal of the ast first, then operate on the results of that which could then be async

src/loaders/css-module-dts-loader/loader.ts
```diff
@@ -0,0 +1,125 @@
...
+	return Promise.all(promises);
+}
+
+const fileType = {
+	css: 'css',
+	ts: 'ts'
+};
+
+export default function (this: Webpack, content: string, sourceMap?: string): Promise<string> {
+	const callback = this.async();
```
matt-gadd: we are given a callback, so does the loader function actually support returning a promise?

src/loaders/css-module-dts-loader/loader.ts
```diff
@@ -0,0 +1,125 @@
...
+	const {
+		type = fileType.ts,
+		instanceName
+	}: {
+		type: string,
+		instanceName: string | void
```
matt-gadd: should this be void? would it not be more likely that `instanceName` is optional?

src/loaders/css-module-dts-loader/loader.ts
```diff
@@ -0,0 +1,125 @@
...
+		return creator.create(filePath, false, true).then((content) => {
+			return content.writeFile();
+		});
+	}
+	else {
+		return Promise.resolve();
+	}
+}
+
+function checkNodeForCSSImport(node: Node, instance?: TSLoaderInstances): Promise<void> {
```
matt-gadd: if you make this function sync like I commented on below, you can then drop all the else blocks for returning resolved promises.

src/loaders/css-module-dts-loader/loader.ts
```diff
@@ -0,0 +1,125 @@
...
+type Webpack = {
+	resourcePath: string;
+	async(): (error: Error | null, result: string, sourceMap?: string) => void;
+}
+
+const creator: DtsCreatorInstance = new DtsCreator();
+
+const mTimeMap = new Map<string, Date>();
+
+function generateDTSFile(filePath: string): Promise<void> {
```
matt-gadd: If you always need to return a Promise, you can start a chain with a `Promise.resolve`, it will neaten the following up to:
```ts
function generateDTSFile(filePath: string): Promise<void> {
  return Promise.resolve().then(() => {
    const { mtime } = statSync(filePath);
    const lastMTime = mTimeMap.get(filePath);

    if (!lastMTime || mtime > lastMTime) {
      mTimeMap.set(filePath, mtime);
      return creator.create(filePath, false, true)
        .then((content) => content.writeFile());
    }
});
```

tomdye: Good call 👍

tomdye:
@matt-gadd all comments addressed

src/loaders/css-module-dts-loader/loader.ts
```diff
@@ -0,0 +1,110 @@
...
+			forEachChild(node, (childNode: Node) => {
+				const path = getCssImport(childNode);
+				path && filePaths.push(path);
+			});
+			break;
+	}
+	return filePaths;
+}
+
+export default function (this: Webpack, content: string, sourceMap?: string) {
```
matt-gadd: again I think you can tidy the flow of this up. because you are always calling `callback` with the same result you should probably only do it once. something like:
```typescript
export default function (this: Webpack, content: string, sourceMap?: string) {
	const callback = this.async();
	const { type = 'ts', instanceName }: { type: string, instanceName?: string } = getOptions(this);

	Promise.resolve().then(() => {
		switch (type) {
			case 'css':
				return generateDTSFile(this.resourcePath);
			break;
			case 'ts':
				const sourceFile = createSourceFile(this.resourcePath, content, ScriptTarget.Latest, true);
				const cssFilePaths = checkNode(sourceFile);

				if (cssFilePaths.length && instanceName) {
					const tsInstances = instances.getTypeScriptInstance({ instance: instanceName });
					tsInstances.instance.files[this.resourcePath] = false;
				}

				const promises = cssFilePaths.map((cssFilePath) => generateDTSFile(cssFilePath));
				return Promise.all(promises);
			}
			break;
		}
	})
	.then(() => callback(null, content, sourceMap));
}
```

tomdye: ok

src/loaders/css-module-dts-loader/loader.ts
```diff
@@ -0,0 +1,110 @@
...
+				callback(null, content, sourceMap);
+			});
+			break;
+		case 'ts':
+			const sourceFile = createSourceFile(this.resourcePath, content, ScriptTarget.Latest, true);
+			const cssFilePaths = traverseNode(sourceFile);
+
+			if (cssFilePaths.length) {
+				if (instanceName) {
+					const tsInstances = instances.getTypeScriptInstance({ instance: instanceName });
```
matt-gadd: this doesn't return instances does it? it returns a wrapper with the instance and/or error right? so would `instanceWrapper` make more sense as a variable name?

matt-gadd: also do you want to do anything with the `e`/error if their is one? (I noticed you had this before)

tomdye: yes it would, I couldn't think of a decent name.

tests/unit/loaders/css-module-dts-loader.ts
```diff
@@ -0,0 +1,262 @@
...
+		});
+		mockFs.statSync.reset();
+
+		return new Promise((resolve, reject) => {
+			loaderUnderTest.call(defaultScope, cssContent);
+			loaderUnderTest.call({
+				async() {
+					return () => {
+						try {
+							assert.isTrue(mockFs.statSync.calledTwice);
```
matt-gadd: this would read a lot nicer if you did the work then the assertion block came after, rather than injecting the assertions

matt-gadd: (and remove some of the nesting)

PR #73, opened 2017-02-23
Remove `cldr-data`, update for compat with `@dojo/i18n` alpha.6

### LINKED ISSUE #71: Update i18n to alpha.6
**Enhancement**

With `@dojo/i18n@v2.0.0-alpha.6`, we can remove `cldr-data` as a dependency. This also includes updating the i18n plugin to extract arrays of URLs passed to `@dojo/i18n/cldr/load.default`. As with use of `@dojo/i18n`, users will be required to provide their own CLDR data, but at the same time are not required to include it if their application does not need it.

With this update, users will be able to load CLDR data as follows:

```typescript
import loadCldrData from '@dojo/i18n/cldr/data';

loadCldrData([
	'cldr-data/main/{locale}/currencies.json',
	'cldr-data/main/{locale}/numbers.json',
	'cldr-data/supplemental/currencyData.json',
	'cldr-data/supplemental/likelySubtags.json',
	'cldr-data/supplemental/numberingSystems.json',
	'cldr-data/supplemental/ordinals.json',
	'cldr-data/supplemental/plurals.json'
]);
```

At build time, the URLs will be extracted and loaded, and any URL with a `{locale}` token will be loaded for both the `defaultLocale` as well as any `supportedLocales`.

mwistrand:

**Description:**

Since `cldr-data` is no longer a dependency of `@dojo/i18n@2.0.0-alpha.6`, it no longer needs to be included as dependency of `cli-build` either. Instead, users are encouraged to load their own CLDR data as they need it. To accommodate this, the build will now parse URL arrays from `@dojo/i18n/cldr/load.default` and ensure their data are injected into the build for all supported locales.

Resolves #71


mwistrand:
I missed that the configuration helper hadn't already been hooked up to read configuration options from an app-level `.dojorc`. I'll add that as well.

src/plugins/I18nPlugin.ts
```diff
@@ -137,14 +112,65 @@ export default class DojoI18nPlugin {
...
+				}
+
+				return callback(null, result);
+			});
+
+			data.normalModuleFactory.plugin('parser', (parser: any) => {
+				parser.plugin('program', (ast: any) => {
+					const path = parser.state.current.userRequest;
+					if (path) {
+						astMap.set(path, ast);
```
mwistrand: Since it cannot be determined whether a module requires `@dojo/i18n/cldr/load` until after it has been parsed, the AST objects are stored in the map indiscriminately. For large applications, I wonder whether the memory consumption would be problematic. However, I have yet to find a reliable means of tracking when a module's dependencies have been completely parsed to remove that module's AST object from the cache. If anyone has any suggestions about managing this, I'd greatly appreciate it.

mwistrand: Actually, if I map dependency mids to their issuers during the `before-resolve` event, and then remove them during the `parser` event, I should be able to track when all dependencies have been parsed.

PR #78, opened 2017-03-01
[fix] remove compose dep as no longer needed.

### LINKED ISSUE #77: We can remove 'compose' dep
[nit] `compose` is no longer used in this repo so we can remove the dep.

rishson:
Remove `compose` dep.
Resolves #77


rishson:
Ugh - wrong branch

PR #81, opened 2017-03-06
Eject command for cli-build

### LINKED ISSUE #58: Implement `eject` for the command configuration
**Enhancement**

`dojo/cli` supports now ejecting configuration for advanced users. The `eject` should be implemented for the build configuration.

rorticus:
**Type:** feature

**Description:**

`eject` command for `cli-build`.  Ejecting the build command will install all the dependencies you need for webpackery.  After the eject, you can run,

```bash
$ node_modules/.bin/webpack --config=config/build-webpack/webpack.config.js
```

Resolves #58


src/main.ts
```diff
@@ -165,6 +169,45 @@ const command: Command = {
 		else {
 			return compile(config(configArgs), options);
 		}
+	},
+	eject(helper: Helper) {
+		const ejectOutput: EjectOutput = {
+			npm: {
+				devDependencies: {
```
mwistrand: Is there any way this data could be pulled from the `dependencies` listed `package.json` (with the exception of `@dojo/cli-build-webpack`)? That way, this list wouldn't have to be maintained in two different places.

src/webpack.config.ts
```diff
@@ -2,24 +2,24 @@ import webpack = require('webpack');
...
 const postcssImport = require('postcss-import');
 const postcssCssNext = require('postcss-cssnext');
 
-import CoreLoadPlugin from './plugins/CoreLoadPlugin';
-import I18nPlugin from './plugins/I18nPlugin';
-import InjectModulesPlugin from './plugins/InjectModulesPlugin';
+const isCLI = process.env.DOJO_CLI;
+const CoreLoadPlugin = isCLI ? require('./plugins/CoreLoadPlugin').default : require('@dojo/cli-build-webpack/plugins/CoreLoadPlugin').default;
+const I18nPlugin = isCLI ? require('./plugins/I18nPlugin').default : require('@dojo/cli-build-webpack/plugins/I18nPlugin').default;
+const InjectModulesPlugin = isCLI ? require('./plugins/InjectModulesPlugin').default : require('@dojo/cli-build-webpack/plugins/InjectModulesPlugin').default;
```
mwistrand: The duplication between the `require`s could be cleaned up with something like:

```typescript
const packagePath = isCLI ? '.' : '@dojo/cli-build-webpack';
const SomePlugin = require(`${packagePath}/plugins/SomePlugin').default;
```

rorticus: Good idea! Updated.

src/main.ts
rorticus: Well either way we are going to have a list to maintain. Granted, a list of excluded dependencies will be shorter (for now) than a list of included dependencies, but I'm not sure if that's just a step sideways. I like the idea of having just the one list, I'm just not quite sure how to pull it off because the `dependencies` section has a handful of packages that aren't in this list (`cldr-data`, `@dojo/shim`/`core`/`i18n`, `has`).

tomdye: good call @mwistrand. Might be worth us looking into loading the `package.json` in here and pulling out versions from named requirements perhaps. For beta I think this will be fine though.

agubler: Personally I think we need to change this for beta.

tomdye: Ok, fair. So we'd need a way to support pulling from package.json OR specifying a specific semver in the case of `cli-` dependencies that aren't in the package.json. @agubler Would you see this as something that the `cli-eject` command should handle, ie. pass the package.json along as part of `eject` output or should each command handle it itself?

agubler: Remind me why we can't take the full list of dependencies?

tomdye: well they're not the same list of dependencies. The cli command has a lot more than the ejected part.

agubler: I don't get what you mean by that?

tomdye:
This PR requires a CLI release to include the `DOJO_CLI` env var

tomdye:
@rorticus Not sure why this is failing CI, but can you update to `@dojo/cli: 2.0.0-alpha.11` please?

tomdye:
Restarted build

src/main.ts
rorticus: So... what do you want me to do here? 😄

tomdye: I don't think we can use the full set of deps from the command's package.json. We likely need to pull it in and pull out the versions for a given subset.

agubler: Okay so I don't know why we cannot just take the all the packages from dependencies in the `package.json` https://github.com/dojo/cli-build/blob/master/package.json#L49

If there are any missing or in the incorrect section (devDep) we should fix that rather than maintaining an external list (without versions mind you) as a work around.

rorticus: @agubler , these dependencies are in the `dependencies` section in `package.json` but are not being outputted as `devDependencies` of the eject command,

```
"@dojo/core": "2.0.0-alpha.20",
"@dojo/has": "2.0.0-alpha.7",
"@dojo/i18n": "2.0.0-alpha.5",
"@dojo/shim": "2.0.0-beta.8",
"cldr-data": "^30.0.1"
```

I agree that realistically, those should probably be output also...

rorticus: OK, updated to use packages from `package.json`!

src/main.ts
```diff
@@ -118,10 +119,12 @@ function compile(config: webpack.Config, options: WebpackOptions): Promise<any>
...
 function buildNpmDependencies(): any {
 	try {
-		const packageJson = JSON.parse(fs.readFileSync(path.join(__dirname, 'package.json')).toString());
+		const packagePath = pkgDir.sync(__dirname);
+		const packageJsonFilePath = path.join(packagePath, 'package.json');
+		const packageJson = <any> require(packageJsonFilePath);
 
 		return {
-			'@dojo/cli-build-webpack': packageJson.version,
+			[packageJson.name]: packageJson.version,
```
tomdye: nice 👍

agubler:
Does eject work if the command is install globally?

rorticus:
Looks like eject worked when installed globally

PR #84, opened 2017-03-08
Adds hint to eject command

tomdye:
**Type:** bug / feature

**Description:**
Add's eject hint output:
screenshot 2017-03-08 15 50 17

Requires CLI release

Resolves #81

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "test": "grunt test"
   },
   "devDependencies": {
-    "@dojo/cli": "2.0.0-alpha.11",
+    "@dojo/cli": "2.0.0-alpha.12",
     "@dojo/interfaces": "^2.0.0-alpha.11",
     "@dojo/loader": "2.0.0-beta.9",
     "@types/chai": "^3.4.34",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -17,7 +17,7 @@
     "test": "grunt test"
   },
   "devDependencies": {
-    "@dojo/cli": "2.0.0-alpha.11",
+    "@dojo/cli": "2.0.0-alpha.12",
```
tomdye: this hasn't been released yet

src/main.ts
```diff
@@ -211,7 +212,10 @@ const command: Command = {
 				files: [
 					'./webpack.config.js'
 				]
-			}
+			},
+			hints: [
+				'to build run ' + underline('webpack -config ./config/build-webpack/webpack.config.js')
```
agubler: assumes webpack installed globally?

tomdye: fair point, isn't that how most people have it set up? I could change it to ref the one in `node_modules` if you prefer?

agubler: does ejecting add webpack to the projects package.json?

tomdye: yes it does

PR #89, opened 2017-03-16
Adding automatic lazy bundling, issue #87

### LINKED ISSUE #87: Support automatic lazy widget bundling
**Enhancement**

When using `@dojo/core/load` to lazy load a module, we want automatically add the module to its own bundle

Given,

```typescript
registry.define('my-widget', () => {
    return load(require, './MyWidget')
        .then(([ MyWidget ]) => MyWidget.default);
});
```

We want `cli-build` to add `MyWidget` to it's own bundle using something like,

```typescript
require(`bundle-loader?lazy&name=${widgetName}!./Widget`);
```

rorticus:
**Type:** feature

**Description:**

This adds automatic bundling for the lazy widget loading pattern.  The pattern looks like this,

```typescript
import load from '@dojo/core/load';

registry.define('some-widget', () => {
    return load(require, './widgets/SomeWidget').then(([someWidget]) => someWidget.default);
});
```

Detecting this pattern will automatically create a new output file that will get loaded dynamically.

*About the pattern*

The pattern being looked for is a method call where the first parameter is a `require` identifier, the second parameter is a `string`, and the entire method call is wrapped inside of a `.define` method call.

You can disable bundling with the `--disableLazyWidgetDetection` option.

Resolves #87


src/plugins/CoreLoadPlugin.ts
```diff
@@ -54,11 +61,91 @@ function stripPath(basePath: string, path: string): string {
...
+	ignoredModules?: string[];
+	basePath?: string;
+}
+
 /**
  * A webpack plugin that forces webpack to ignore `require` passed as a value, and replaces `@dojo/core/load` with a
  * custom function that maps string module IDs to webpack's numerical module IDs.
  */
 export default class DojoLoadPlugin {
+	detectLazyLoads: boolean;
```
agubler: Are these properties `private`?

src/plugins/CoreLoadPlugin.ts
```diff
@@ -54,11 +61,91 @@ function stripPath(basePath: string, path: string): string {
...
+	lazyChunkNames: DojoLoadChunkNames;
+	ignoredModules: Set<string>;
+
+	constructor(options: DojoLoadPluginOptions = {}) {
+		const { detectLazyLoads, chunkNames, ignoredModules, basePath = '' } = options;
```
agubler: If you default the options here you won't need the `||` when initialising the properties.

src/plugins/CoreLoadPlugin.ts
```diff
@@ -54,11 +61,91 @@ function stripPath(basePath: string, path: string): string {
...
+
+		this.detectLazyLoads = detectLazyLoads || false;
+		this.lazyChunkNames = chunkNames || {};
+		this.ignoredModules = new Set<string>();
```
agubler: Can initialise this inline in the property declaration (if you wanted).

src/plugins/CoreLoadPlugin.ts
```diff
@@ -54,11 +61,91 @@ function stripPath(basePath: string, path: string): string {
...
+
+		if (ignoredModules) {
+			ignoredModules.forEach(moduleName => {
+				const absolutePath = path.resolve(basePath, moduleName);
+
+				this.ignoredModules.add(absolutePath);
+			});
+
+			console.log(this.ignoredModules);
```
agubler: Did we mean to log the ignored modules here?

src/plugins/util/parser.ts
```diff
@@ -0,0 +1,159 @@
...
+ * Determine whether the specified node is shadowing any of the specified variable names.
+ *
+ * @param item
+ * The node to test
+ *
+ * @param importNames
+ * A list of variable names.
+ *
+ * @return
+ * `true` if the
```
agubler: if the..... it's a mystery :)

src/plugins/util/parser.ts
```diff
@@ -0,0 +1,159 @@
...
+ * Determine whether the provided node is a function declaration.
+ *
+ * @param item
+ * The item to test.
+ *
+ * @return
+ * `true` if the node represents either function declaration or a function expression.
+ */
+export function isFunctionDefinition(item: any): item is BaseFunction {
+	const { type } = item;
```
agubler: Can `item` be undefined? you have guarded against it in `isArrayExpression`

src/plugins/util/parser.ts
```diff
@@ -0,0 +1,159 @@
...
+		'BlockStatement',
+		'ForInStatement',
+		'ForOfStatement',
+		'ForStatement',
+		'FunctionDeclaration',
+		'FunctionExpression',
+		'LetStatement',
+		'Program',
+		'TryStatement'
+
```
agubler: line break?

src/plugins/util/parser.ts
```diff
@@ -0,0 +1,159 @@
...
+	].forEach((type: string) => {
+		nextItemMap[ type ] = getters.body;
+	});
+
+	return function (item: any): any {
```
agubler: Can `item` be undefined?

src/plugins/CoreLoadPlugin.ts
```diff
@@ -72,17 +159,141 @@ export default class DojoLoadPlugin {
...
+												const edep = new RequireEnsureItemDependency(second.value, second.range);
+												edep.loc = dep.loc;
+												dep.addDependency(edep);
+											});
+
+											/*
+											 By default, the require.ensure is not going to execute when we want. We wrap it in a function block
+											 to control the execution.
+											 */
+											(<any> old).addDependency(new ConstDependency('function() { return (', fnExpression.range[0]));
```
agubler: As you have to cast all the usages would it better to `const old: any = parser.state.current;` (if that works) and then just cast back to the current type when it gets reassigned? `parser.state.current = <CurrentType> old;`?

src/plugins/CoreLoadPlugin.ts
```diff
@@ -1,10 +1,17 @@
 import * as path from 'path';
+import { getBasePath, resolveMid } from './util/main';
+import { CallExpression, Program } from 'estree';
+import { getNextItem } from './util/parser';
+import Set from '@dojo/shim/Set';
```
agubler: As this runs in Node do we need the shim?

src/plugins/CoreLoadPlugin.ts
rorticus: Nope... oops 😄

src/plugins/CoreLoadPlugin.ts
rorticus: Removed the `<any>` casts everywhere for `old`

src/plugins/util/parser.ts
rorticus: Yup!

src/plugins/util/parser.ts
rorticus: Item shouldn't be undefined unless something is wrong with the AST tree, but I added a check regardless.

src/plugins/util/parser.ts
rorticus: Item shouldn't be undefined unless something is wrong with the AST tree, but I added a check regardless.

src/plugins/util/parser.ts
rorticus: Haha.

rorticus:
The tests dont cover the webpack config at all, so I manually tested the population of `ignoredModules` from the `bundles` config.

Widget loading,

```typescript
registry.define('todo-footer', () => {
	return load(<any> require, './widgets/TodoFooter').then(([ todoFooter ]) => todoFooter.default);
});

registry.define('view-chooser', () => {
	return load(<any> require, './widgets/ViewChooser').then(([ viewChooser ]) => viewChooser.default);
});

registry.define('filters', () => {
	return load(<any> require, './widgets/TodoFilter').then(([ todoFilter ]) => todoFilter.default);
});
```

`.dojorc`:

```json
{
	"build-webpack": {
		"bundles": {
			"widgets": [
				"src/widgets/ViewChooser",
				"src/widgets/TodoFilter"
			]
		}
	}
}
```

`dojo build` then generates the following files,

* `main.js`
* `widgets.js` (based on bundle information in `.dojorc`)
* `TodoFooter.js` (based on automatic detection)

rorticus:
Added a check for `@dojo/core/load` before processing the rest of the AST. Should speed things up a bit.

src/plugins/CoreLoadPlugin.ts
mwistrand: I had to do the same thing with `Map` in `I18nPlugin` due to type collisions in what was either an update to TS or to `@dojo/shim` (I can't remember exactly which).

PR #92, opened 2017-03-16
Adding manual bundling from .dojorc, issue #90

### LINKED ISSUE #90: Add manual file bundling
**Enhancement** 

Given a `.dojorc` like this:

```
{
    'build': {
        'bundles': {
             'foo-bundle': [
                 './src/todo/Foo',
                 './src/something/Else'
             ]
        }
    }
}
```

Create a bundle `foo-bundle` consisting of the given files.

rorticus:
**Type:** bug / feature

**Description:**

Manual lazy loading bundling using `.dojorc`.  `.dojorc` looks like this:

```json
{
	"build-webpack": {
		"bundles": {
			"widgets": [
				"src/widgets/TodoFooter",
				"src/widgets/ViewChooser",
				"src/widgets/TodoFilter"
			]
		}
	}
}
```

This will create a `widgets.js` file containing `TodoFooter`, `ViewChooser`, and `TodoFilter`.

Resolves #90

### DIFF
--- a/src/webpack.config.ts
+++ b/src/webpack.config.ts
@@ -179,7 +179,8 @@ function webpackConfig(args: Partial<BuildArgs>) {
 			}, () => {
 				return path.resolve('./dist');
 			}),
-			filename: '[name].js'
+			filename: '[name].js',
+			chunkFilename: '[name].js'
 		},
 		devtool: 'source-map',
 		resolve: {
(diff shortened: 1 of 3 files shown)


rorticus:
I wanted to get this in here. I'll work on getting some test coverage on this.

src/webpack.config.ts
```diff
@@ -237,6 +238,25 @@ function webpackConfig(args: Partial<BuildArgs>) {
 					return [
 						{ test: /custom-element\.js/, loader: `imports-loader?widgetFactory=${args.element}` }
 					];
+				}),
+				...includeWhen(args.bundles && Object.keys(args.bundles).length, () => {
+					const loaders: any[] = [];
+
+					Object.keys(args.bundles).forEach(bundleName => {
+						(args.bundles || {})[ bundleName ].forEach(fileName => {
```
agubler: Surprised you need to default `args.bundles` to `{}`, would have thought `Object.keys` would fail is it was undefined.

rorticus: I've run into this problem with TS before. I could probably have put `args.bundles!` as well. It seems like the type inference for variables doesn't extend into closures, as TS seems to forget that `args.bundles` wasn't null a second ago :)

Removing the `|| {}` causes a "variable could be undefined" error. Any thoughts?

PR #102, opened 2017-03-22
Processing `.m.css` instead of `.css`

rorticus:
**Type:** feature

**Description:**

Modifies the build command to process `.m.css` and leave `.css` files alone.

Resolves https://github.com/dojo/widgets/issues/76

### DIFF
--- a/src/webpack.config.ts
+++ b/src/webpack.config.ts
@@ -111,7 +111,7 @@ function webpackConfig(args: Partial<BuildArgs>) {
 			};
 		}),
 		plugins: [
-			new NormalModuleReplacementPlugin(/\.css$/, result => {
+			new NormalModuleReplacementPlugin(/\.m.css$/, result => {
 				const requestFileName = path.resolve(result.context, result.request);
 				const jsFileName = requestFileName + '.js';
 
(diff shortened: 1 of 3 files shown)


src/webpack.config.ts
```diff
@@ -95,31 +95,31 @@ function webpackConfig(args: Partial<BuildArgs>) {
 		}, args => {
 			return {
 				'src/main': [
-					path.join(basePath, 'src/main.css'),
+					path.join(basePath, 'src/main.m.css'),
```
tomdye: Do you think that `main.css` should be a module? The issue I had that initially raised this concern around naming was that I wanted to import some vendor css in my `main.css` file and could not stop it from turning it all into a `css-module` and breaking everything. My personal thought is that `main.css` should *not* be a `.m.css` file.

src/webpack.config.ts
```diff
@@ -130,7 +130,7 @@ function webpackConfig(args: Partial<BuildArgs>) {
 			}),
 			includeWhen(args.element, () => {
 				return new CopyWebpackPlugin([
-					{ context: 'src', from: '**/*', ignore: [ '*.ts', '*.css', '*.html' ] }
+					{ context: 'src', from: '**/*', ignore: [ '*.ts', '*.m.css', '*.html' ] }
```
tomdye: Do we not want to ignore all `.css` ?

src/webpack.config.ts
rorticus: Updated to not be a `.m.css` file!  It does make sense that we won't want the "main" css file to be a css module.

src/webpack.config.ts
rorticus: Updated.

PR #104, opened 2017-03-23
Add build option `target` to change webpack target between "web" and …

edhager:
The builder is not setting Webpack's `target` option and the default is "web" which builds chunks for the browser environment.  Using `@dojo/core/load` causes  Webpack to inject code into the chunks that assumes the global `window` exists.  This fails when the chunks are loaded in a node.js environment.

This PR adds a `target` option to the builder which maintains the "web" target by default.

Fix dojo/cli-test-intern#17

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -205,6 +205,11 @@ const command: Command = {
 			describe: 'Disable lazy widget loading detection',
 			type: 'boolean'
 		});
+
+		options('target', {
+			describe: 'Build target environment "web" (default) or "node"',
+			type: 'string'
+		});
 	},
 	run(helper: Helper, args: BuildArgs) {
 		const dojoRc = helper.configuration.get('build-webpack') || Object.create(null);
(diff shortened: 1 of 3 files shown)


edhager:
This is fix 1 of 2 for dojo/cli-test-intern#17.  The second fix is in @dojo\cli-test-intern.

matt-gadd:
For now i'd recommend dropping this option completely and just hardcoding it to web in the webpack config. like other options if they want to change something this fundamental they can eject the command. The mantra behind the cli is to offer as little configuration as needed, and currently we don't support SSR (and even/when we do its likely to be under jsdom with likely a window implementation anyway).

matt-gadd:
Seeing actually how this works, I realise we are now passing `node` from the test command? Does this mean we are testing a bundle in one environment and then shipping it to another? Is this a good idea?

agubler:
We have gone with not code splitting for the test bundle (which is used for running unit tests) which has been addressed https://github.com/dojo/cli-test-intern/pull/26

PR #124, opened 2017-04-21
Include optimize-css-assets-webpack-plugin for dist build.

### LINKED ISSUE #123: Optimize CSS output for dist builds
**Enhancement** 

We currently do not apply any CSS optimization to our build output so we have some duplicated classes (due to css-module composition) and unused variable blocks as well as whitespace and non-shorthand style definitions.
Adding the `optimize-css-assets-webpack-plugin` webpack plugin for dist builds will achieve this for us.

**Example code**
``` typescript
...includeWhen(!args.watch && !args.withTests, (args) => {
   return [new OptimizeCssAssetsPlugin()( options )];
}),
```

tomdye:
**Type:** bug / feature

**Description:**
Adds `OptimizeCssAssetsPlugin` for dist mode.

Resolves #123

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
     "imports-loader": "^0.7.0",
     "json-css-module-loader": "^1.0.0",
     "loader-utils": "^1.0.2",
+    "optimize-css-assets-webpack-plugin": "^1.3.1",
     "pkg-dir": "^1.0.0",
     "postcss-cssnext": "^2.9.0",
     "postcss-import": "^9.0.0",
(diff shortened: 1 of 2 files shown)


tomdye:
This currently removes the CSS sourcemaps as well. Do we want these for our built CSS?

mwistrand:
This build is failing to due to a breaking change in one of webpack's modules used by `CoreLoadPlugin`. There is a fix included in #122, but the build failure is unrelated to any functionality included in this PR.

tomdye:
@mwistrand im just going to have a quick look to see if i can turn sourcemaps on, this `cssProcessorOptions: { sourcemap: true }` should do it.

mwistrand:
@tomdye Am I correct that `cssnano` does not provide an option for external source maps? So far I only see inline source map support, which, as you pointed out, adds a lot of weight to the built CSS file.

PR #138, opened 2017-05-10
Allow external CSS modules to be imported.

### LINKED ISSUE #112: Can't load CSS modules from node modules
**Bug** 

It looks like you can't build code that imports CSS modules that aren't in the `src/` directory.  I _think_ it's a problem with `src/loaders/css-module-dts-loader/loader.ts#getCssImport`.

This was discovered while looking into https://github.com/dojo/meta/issues/138 .

**Code**

```typescript
import * as localStyles from './some/local/file.m.css';
import * as foreignStyles from '@dojo/widgets/common/styles/base.m.css';
```

You can error when running `dojo build`.

**Expected behavior:**

I expect it to build 😄 

**Actual behavior:**

```
(node:28421) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: ENOENT: no such file or directory, stat '/Users/rorymulligan/Desktop/dojo2/examples/todo-mvc/src/@dojo/widgets/common/base.m.css
```

mwistrand:
**Type:** bug

**Description:**

Updates the `css-module-dts-loader` loader to allow CSS modules outside of `src/` to be loaded. With this any, any path that does not begin with a dot (`.`) is assumed to be either an absolute path or mid.

Also moves the `util/` modules from `src/plugins/util/` to `src/util/` since they are no longer used only by or only useful to plugins.

Resolves #112


kitsonk:
Closing in favour of #178

PR #151, opened 2017-05-23
Fix: use correct banner.md path when ejecting.

### LINKED ISSUE #145: Eject no longer works
**Bug**

Ejecting the webpack config no longer works. It looks like https://github.com/dojo/cli-build/pull/119 's usage of `__dirname` over `packagePath` is responsible looking at the error.

mwistrand:
**Type:** bug

**Description:**

When ejecting, use the correct file path for the banner used by the webpack banner plugin so that the build can pass without "file not found" errors. Verified with a local app created with `$ dojo create`.

Resolves #145


mwistrand:
@kitsonk Currently the webpack config has no tests, in part because it resolves all paths and instantiates all plugins on load. If we wanted to include tests for it, we could perhaps try exporting a `createConfig` method that encapsulates 100% of the logic, and then change the default export to `export default createConfig()`;

kitsonk:
Or use [`vm.runInContext`](https://nodejs.org/api/vm.html#vm_vm_runincontext_code_contextifiedsandbox_options) to stub the necessary context and evaluate the result is as expected.  No changes then to the structure of the code.

mwistrand:
Thanks, @kitsonk. I've been working on a test for this, but am hung up on `TypeError: require.resolve is not a function` errors. Passing in `require` works fine, but it cannot register the existence of `require.resolve` (even when stubbing out `require`). Based on how TS constructs its UMD wrapper, at the very least the following must be passed in:

```typescript
const exports = {};
const context = vm.createContext({
    exports,
    module: { exports },
    require
});
```

On top of that, I've needed to jump through additional hoops to ensure that modules are loaded correctly (`./` in some cases, `@dojo/cli-build-webpack` with a falsy `DOJO_CLI` env flag). If you have any suggestions, I'd appreciate hearing them.

kitsonk:
Assuming you are running under the Dojo 2 loader for intern tests, don't you want `require.nodeRequire` to be passed in as `require` into a context?  I believe using [`mockery`](https://www.npmjs.com/package/mockery) at that point should allow you to mock any other requirements that you want/need to stub out.

matt-gadd:
can you explain the separation of `Sandbox` and `MockModule` a little? I'm struggling to get the intent (ie a lot of "mocking" is going on in `Sandbox` now?). If there is no real reason behind the separation can `MockModule` just be removed?

src/webpack.config.ts
```diff
@@ -18,6 +18,7 @@ const packagePath = isCLI ? '.' : '@dojo/cli-build-webpack';
 const CoreLoadPlugin = require(`${packagePath}/plugins/CoreLoadPlugin`).default;
 const I18nPlugin = require(`${packagePath}/plugins/I18nPlugin`).default;
 const basePath = process.cwd();
+const bannerPath = isCLI ? path.join(__dirname, './banner.md') : require.resolve(`${packagePath}/banner.md`);
```
matt-gadd: can this not just be `packagePath`? which is adjusted accordingly based on `isCLI` anyway?

mwistrand:
@matt-gadd `MockModule` actually mocks the module being tested, but that's not what these tests need to do. `Sandbox` is perhaps a misleading name, and perhaps I'm just overthinking this, but I removed the sandboxing/dep mocking into a separate module since that's all I needed for these tests.

src/webpack.config.ts
mwistrand: Ah, ```require.resolve(`${packagePath}/banner.md`)``` should work in both scenarios.

mwistrand:
@matt-gadd I ended up removing `Sandbox` in favor of adding a `start` method directly to `MockModule`. `MockModule` in this case is still not being used to mock the config, but at least we have only one module to deal with when writing tests.

PR #167, opened 2017-06-23
Changes to get `composes` working with css-modules and our build system

### LINKED ISSUE #166: Dojo Build fails when using `composes` with local files
**Bug** 

When attempting to use `composes` with a local .css file import the cli build command will currently fail. This appears to be because the css-module composes functionality adds `require` statements to the generated webpack output that we did not account for in our `css-module-decorator-loader` when adding our theme key.

We need to re-visit this to allow for the extra requires and cannot assume the provided format is `JSON` as before.

The output in fact, can look like this:
```
 "root": "_3kYHOfq9 " + require("-!../../../node_modules/css-loader/index.js?modules&sourceMap&importLoaders=1&localIdentName=[hash:base64:8]!../../../node_modules/postcss-loader/index.js?{\"plugins\":[null,{\"version\":\"5.2.17\",\"plugins\":[null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,null,
null,null,null,null,null,null,null,null,null,null,null],\"postcssPlugin\":\"postcss-cssnext\",\"postcssVersion\":\"5.2.17\"}]}!./base.css").locals["bob"] + "",
        "text": "_3WTyNu_u"
```

tomdye:
**Type:** bug / feature

- updates postcss dependencies
- adds postcss.config file
- adds test for non json-format exports (composes output)

Resolves #166


src/postcss.config.ts
```diff
@@ -0,0 +1,12 @@
+module.exports = {
```
kitsonk: Untested... we should have even a simple validation check for something like this.

src/postcss.config.ts
```diff
@@ -0,0 +1,12 @@
...
+	plugins: [
+	require('postcss-import')(),
```
kitsonk: Indentation alignment...

package.json
```diff
@@ -65,9 +65,9 @@
     "loader-utils": "^1.0.2",
     "optimize-css-assets-webpack-plugin": "^1.3.1",
     "pkg-dir": "^1.0.0",
-    "postcss-cssnext": "^2.9.0",
-    "postcss-import": "^9.0.0",
-    "postcss-loader": "^1.3.0",
+    "postcss-cssnext": "2.11.0",
```
kitsonk: Was there a reason for pinning the dependencies?

src/postcss.config.ts
kitsonk: Test maybe added, but it is still being reported as uncovered... 😖

package.json
tomdye: I seem to remember that @agubler asked me to

tomdye:
@kitsonk what do you want me to do re. the file showing as uncovered even though it's now clearly tested?

kitsonk:
> what do you want me to do re. the file showing as uncovered even though it's now clearly tested?

Find out why it isn't being covered?  It is being instrumented, but the tests don't actually seem to be covering it. https://codecov.io/gh/dojo/cli-build/pull/167/diff

agubler:
@tomdye can you resolve the conflicts please 👍

kitsonk:
Also, let's not worry about the coverage for this too... so if we rebase than I would be comfortable with a merge.

PR #168, opened 2017-06-28
Service worker support

maier49:
**Type:** bug / feature

**Description:**

Relates to dojo/meta#169. Adds a flag `--serviceWorker`. If this flag is specified, and not running in watch mode or building tests, the `process.env.DOJO_SERVICE_WORKER` flag will be set to true, and the `sw-precache-webpack-plugin` will be enabled, adding support for caching all resources with a service worker to the app.  

The [hacker news PWA PR](https://github.com/dojo/examples/pull/227), which is still a WIP, if built with this branch will work offline. With this and the changes it depends on in `dojo/stores` for IndexedDB the entire app currently runs offline after the first load.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,6 @@
     "mockery": "^1.7.0",
     "remap-istanbul": "^0.6.4",
     "sinon": "^1.17.5",
-    "sw-precache-webpack-plugin": "^0.11.3",
     "yargs": "^5.0.0"
   },
   "dependencies": {
@@ -71,6 +70,7 @@
     "postcss-loader": "1.3.3",
     "source-map-loader-cli": "0.0.1",
     "style-loader": "0.13.2",
+    "sw-precache-webpack-plugin": "^0.11.3",
     "ts-loader": "2.3.1",
     "tslint": "4.5.1",
     "tslint-loader": "3.5.3",
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "mockery": "^1.7.0",
     "remap-istanbul": "^0.6.4",
     "sinon": "^1.17.5",
+    "sw-precache-webpack-plugin": "^0.11.3",
     "yargs": "^5.0.0"
   },
   "dependencies": {
--- a/src/main.ts
+++ b/src/main.ts
@@ -23,7 +23,7 @@ export interface BuildArgs extends Argv {
 	withTests: boolean;
 	debug: boolean;
 	disableLazyWidgetDetection: boolean;
-	serviceWorkers: boolean;
+	serviceWorker: boolean;
 	bundles: Bundles;
 }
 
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -46,6 +46,7 @@
     "mockery": "^1.7.0",
     "remap-istanbul": "^0.6.4",
     "sinon": "^1.17.5",
+    "sw-precache-webpack-plugin": "^0.11.3",
```
matt-gadd: Shouldn't this be in `dependencies `?

src/main.ts
```diff
@@ -205,6 +206,11 @@ const command: Command = {
 			describe: 'Disable lazy widget loading detection',
 			type: 'boolean'
 		});
+
+		options('serviceWorker', {
```
matt-gadd: Can we move this to a `dojorc` config option instead.

src/webpack.config.ts
```diff
@@ -116,6 +122,9 @@ function webpackConfig(args: Partial<BuildArgs>) {
 					result.request = result.request.replace(/\.m\.css$/, '.m.css.js');
 				}
 			}),
+			new DefinePlugin({
+				'process.env.DOJO_SERVICE_WORKERS': JSON.stringify(includeServiceWorker)
```
matt-gadd: Do we have a standard for env options? Should we put them all under one dojo namespace?

package.json
maier49: Yes, fixed

src/main.ts
maier49: What does that consist of specifically? The tests for dojorc that I've seen all involve loading arguments defined here from a configuration file.

src/webpack.config.ts
maier49: I didn't see anywhere else where we were using environment variables. Putting them all under a dojo namespace makes sense to me. `process.env.DOJO.SERVICE_WORKER`?

src/main.ts
maier49: I removed the call to `options()` so this will now only be supported via `dojorc`. As a follow up to that, it doesn't seem like we have anywhere to document non command line argument configuration options. Would adding a section to the README work?

src/webpack.config.ts
kitsonk: We would need to move `DOJO.CLI` as well...  But agreed.

maier49:
I addressed the comments here, with the exceptions that I didn't change DOJO_CLI because it's not set here, and I haven't tackled the more general testing issue.

maier49:
@kitsonk @matt-gadd All the feedback here has been addressed and it's been rebased against the latest.

agubler:
@maier49 apologies seems we have conflicts again, can we resolve them so we can get this merged. Thanks!

maier49:
In light of the concerns mentioned in [this comment](https://github.com/dojo/meta/issues/83#issuecomment-321250186), we will take a different approach for supporting service workers and other PWA features. It will involve using `.dojorc` and generating appropriate files in `cli-build-webpack`, as opposed to adding code to `cli-create-app`  that may not apply in all cases.

PR #169, opened 2017-06-29
Add worker-loader dependency to support web workers

maier49:
**Type:** feature

* [ ] All code matches the [style guide](https://github.com/dojo/meta/blob/master/STYLE.md)

**Description:**

Adding this dependency would allow the use of webworkers in a dojo app using this syntax
```typescript
const worker: Worker = new (require('worker-loader?name=hackerNewsWebWorker.js!./hackerNewsWebWorker'));
```

It works with a normal build or the dev server, and causes a separate compilation for the webworker. Dependencies can be pulled into the webworker and will be built into a separate bundle.

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
     "umd-compat-loader": "^2.0.1",
     "webpack": "^2.2.1",
     "webpack-bundle-analyzer-sunburst": "^1.2.0",
-    "webpack-dev-server": "^2.3.0"
+    "webpack-dev-server": "^2.3.0",
+    "worker-loader": "^0.8.0"
   }
 }


kitsonk:
Awesome!

Are there any changes to authoring tests in downstream applications?

Also, we should have at least some sort of test with this as well.

matt-gadd:
Although I have no issue with the dependency, we so far have tried to avoid inline webpack loader requires, for the same reasoning as AMD plugins.

kitsonk:
> we so far have tried to avoid inline webpack loader requires

Yeah, good point, I hadn't appreciated that.  Would a better approach be an abstraction that is part of Dojo 2 (maybe core) that ensure that it worked.

Also, if we do go with an approach like this, we should still find a better way to integrate it.  We _can_ type plugins now with something like:

```ts
declare module 'worker-loader?*' {
  export = /* ... */
}
```

Would we not be able to create something that wrapped ES `import()` that we could also inject some feature detection, so the code was isomorphic on Node.js as well?

maier49:
This feature needs further discussion to decide on the best approach. https://github.com/dojo/meta/issues/190 has been created for that purpose.

PR #170, opened 2017-06-29
Copy manifest to output dir

maier49:
**Type:** feature

**Description:**

This is the other part of https://github.com/dojo/cli-create-app/pull/86/files. Addresses dojo/meta#170

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,6 @@
     "mockery": "^1.7.0",
     "remap-istanbul": "^0.6.4",
     "sinon": "^1.17.5",
-    "sw-precache-webpack-plugin": "^0.11.3",
     "yargs": "^5.0.0"
   },
   "dependencies": {
@@ -71,6 +70,7 @@
     "postcss-loader": "^1.3.0",
     "source-map-loader-cli": "^0.0.1",
     "style-loader": "^0.13.1",
+    "sw-precache-webpack-plugin": "^0.11.3",
     "ts-loader": "^2.0.0",
     "tslint": "^4.5.1",
     "tslint-loader": "^3.4.3",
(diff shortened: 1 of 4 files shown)


matt-gadd:
what happens if it's not there? does the `CopyWebpackPlugin` handle it gracefully?

maier49:
@matt-gadd If `manifest.json` doesn't exist it still performs the rest of the build, but there's an error for the failed copy and it is reported as a failed build.

kitsonk:
Wouldn't it be better than to be explicit at the CLI that this is a PWA and that this would be a terminal error, or there is some other way of auto-detecting if we are dealing with a PWA?  Having random errors that should be ignored, unless of course they shouldn't isn't the best from a usability perspective, IMO.

maier49:
Well it's not a PWA if service workers are not enabled, but using service workers doesn't mean it is a PWA. I feel like copying this only when service workers are enabled would be good enough. If not though, we could have a pwa cli argument, that enables this and service workers.

kitsonk:
Yeah, sorry, obviously some way of enabling Service Workers/detecting Service Workers being enabled. 👍

maier49:
@kitsonk Ok this is only adding the copy plugin for the manifest.json file if service workers are enabled now. The easiest way to do that was to pull in the service worker changes though so this now depends on that PR.

maier49:
In light of the concerns mentioned in [this comment](https://github.com/dojo/meta/issues/83#issuecomment-321250186), we will take a different approach for supporting service workers and other PWA features. It will involve using `.dojorc` and generating appropriate files in `cli-build-webpack`, as opposed to adding code to `cli-create-app`  that may not apply in all cases.

PR #172, opened 2017-07-03
Add static has build optimization

### LINKED ISSUE #142: @dojo/has Plugin
**Enhancement** 

Part of the original vision for `@dojo/has` was that as part of the build process, code could be built out by providing a set of static flags to assert during the build process removing the run-time evaluation of the feature flag and instead substituting the build time assertion.

Therefore the `HasPlugin` would take something like the below:

```js
import has from '@dojo/has/has';

if (has('foo')) {
  console.log('foo');
}
else {
  console.log('bar');
}
```

And where the feature `'foo'` would be set to `true` at build time would generate:

```js
import has from '@dojo/has/has';

if (true) {
  console.log('foo');
}
else {
  console.log('bar');
}
```

Which would during minification get rewritten as:

```js
console.log('foo');
```

Also, from a consumer API, we should provide an abstraction that allows the user to build without having to understand the details of the flags.  We would provide several pre-configured sets of flags which would be reduced to the final set being passed to the plugin.  From an end users perspective it would operate something like this:

```sh
$ dojo build --target=ie11
$ dojo build --target=ie11,es2015
$ dojo build --target=touch,es2016
```

kitsonk:
**Type:** feature

**Description:**

This PR allows build optimization using the `has` feature detection library.  During the build process, any static features that have been asserted are used to replace call expressions that look something like `has('feature')`.  For example, if feature `foo` was set to `true` as a static flag, this code:

```ts
import has from `@dojo/has/has`;

if (has('foo')) {
  console.log('has foo');
}
else {
  console.log('does not have foo');
}
```

Would be transformed to this:

```ts
import has from `@dojo/has/has`;

if (true) {
  console.log('has foo');
}
else {
  console.log('does not have foo');
}
```

Then further during the build process, if the build is being minified, uglify will detect the dead code and rewrite it further to something like:

```ts
console.log('has foo');
```

At the command line, users are abstracted from individual flags, and instead supply a set of feature flags they wish to use.  The supported sets are located in `src/features/*.json`.  These are specified by the user by setting the `-f` or `--feature` flag on the CLI.  Multiple feature sets are combined to produce the largest set where the values do not conflict.  For example `es6-promise` is `true` on `chrome` but `false` on `ie11`.  If `chrome` and `ie11` were specified, the values would conflict and this feature would not be statically bound.

A user supplying multiple platforms would look something like this:

```bash
$ dojo build -f chrome safari firefox
```

Which would produce a build that is statically built with features that are common between Chrome, Safari, Firefox.

At the moment, most of the shim library isn't written to have polyfill code within `has()` guard blocks, therefore it doesn't get removed from the build when the native feature is statically expressed.  We will need to revisit those modules in order to create something that builds optimally.

**Todo:**

* [x] - test cases
* [x] - readme/documentation

Resolves #142


kitsonk:
Two thoughts on this...

* In the original issue, I suggested `-t` and `--target`.  I am not sure which makes overall better sense.  Or even something else?
* There is no support for user defined flags.  I think that is fairly critical, but I suggest we tackle that as a separate issue/PR after we land this one.  I think using `.dojorc` to specify custom features which will be _mixed in_ to any other targets makes the most sense.

README.md
```diff
@@ -79,6 +80,33 @@ If the source file does not follow the pattern `create[custom element]Element`,
 dojo build webpack --element=src/path/to/element.ts --elementPrefix=the-special
 ```
 
+### Feature optimization
+
+This command supports the ability to optimize code based on statically asserted features.  The tool can search the source code for modules that attempt to detect features using a [`@dojo/has`](https://github.com/dojo/has) type of API.  By supplying a feature set (or sets) on the command line, the build will be optimized to optimize code branches, making the code smaller and slightly more efficient.  This allows targeting of particular platforms.  The tool will not optimize code, when specifying multiple feature sets, they do not align.  It will instead continue to leave that feature to be detected at run-time.
```
dylans: the build will be optimized to optimize code branches, making the code smaller and slightly more efficient. -> the build will optimize code branches, making the code smaller and more efficient.

dylans: Move this to a separate paragraph and change "The tool will not optimize code, when specifying multiple feature sets, they do not align.  It will instead continue to leave that feature to be detected at run-time." to "When specifying multiple feature sets, if they do not align, the tool will not optimize the source code for these feature sets and will instead continue to leave that feature to be detected at run-time."

README.md
```diff
@@ -79,6 +80,33 @@ If the source file does not follow the pattern `create[custom element]Element`,
...
+|`android`|This feature set represents Android 5+ with integrated Chrome browser.  *Note* it is not suitable for Android 4.4.|
+|`chrome`|This feature set represents Chrome 59+ or Opera 46+[<sup>1</sup>](#note-1)|
+|`edge`|This feature set represents Edge 15+[<sup>1</sup>](#note-1)|
+|`firefox`|This feature set represents Firefox 54+[<sup>1</sup>](#note-1)|
+|`ie11`|This feature set represents Internet Explorer 11|
+|`ios`|This feature set represents iOS 10.3+[<sup>2</sup>](#note-2)|
+|`node`|This feature set represents Node.js 7+[<sup>2</sup>](#note-2)|
+|`safari`|This feature set represents Safari 10+[<sup>2</sup>](#note-2)|
+
+<span id="note-1">[1]:</span> Many of these features were present in earlier versions, but the specific version was the GA release at the time of writing when this was validated.
```
dylans: Maybe also say something to the effect that this isn't specific to Chrome 59+ or Opera 46+, and also supports browsers built on the same codebase such as Brave, Vivaldi, and perhaps Electron?

README.md
```diff
@@ -79,6 +80,33 @@ If the source file does not follow the pattern `create[custom element]Element`,
 dojo build webpack --element=src/path/to/element.ts --elementPrefix=the-special
 ```
 
+### Feature optimization
+
+This command supports the ability to optimize code based on statically asserted features.  The tool can search the source code for modules that attempt to detect features using a [`@dojo/has`](https://github.com/dojo/has) type of API.  By supplying a feature set (or sets) on the command line, the build will be optimized to optimize code branches, making the code smaller and slightly more efficient.  This allows targeting of particular platforms.  The tool will not optimize code, when specifying multiple feature sets, they do not align.  It will instead continue to leave that feature to be detected at run-time.
+
+From the command line, the feature sets are provided to the `-f` or `--feature` argument.  The available feature sets are aligned to platforms.  The currently available feature sets are:
+
+|Feature Set|Description|
```
dylans: People may not be impressed to learn that we have what they might think of a browser sniff testing rather than feature testing, so perhaps it's worth explaining a bit more how these feature test lists were created?

src/getFeatures.ts
```diff
@@ -0,0 +1,65 @@
...
+	if (!features.every((exists) => !!exists)) {
+		features.forEach((exists, idx) => {
+			if (!exists) {
+				console.log('Cannot resolve feature set:', featureNames[idx]);
+			}
+		});
+		return {};
+	}
+
+	// Reduce the array of loaded features to the largest set of features where the values don't
```
dylans: don't -> do not

src/plugins/HasPlugin.ts
```diff
@@ -0,0 +1,115 @@
...
+	constructor(features: StaticHasFeatures) {
+		this._features = features;
+	}
+
+	public apply(compiler: Compiler) {
+		// copy features to a local scope, because `this` gets weird
+		const features = this._features;
+		const dynamicFlags = new Set<string>();
+
+		// setup the dependencies for the subtitution
```
dylans: subtitution -> substitution

src/plugins/HasPlugin.ts
```diff
@@ -0,0 +1,115 @@
...
+				console.log();
+			}
+			callback();
+		});
+
+		// we need to hook the compiler
+		compiler.plugin('compilation', (compilation, data) => {
+			// and we want to hook the parser
+			data.normalModuleFactory.plugin('parser', (parser) => {
+				// we need direct access to the AST to properly figure out the has subtitution
```
dylans: subtitution -> substitution

src/plugins/util/walk.ts
```diff
@@ -0,0 +1,60 @@
...
+
+	if (leave) {
+		leave(node, parent, prop, index);
+	}
+}
+
+/**
+ * Walk a estree AST tree
+ * @param node The root estree Node to start walking
+ * @param param1 An object with an `enter` and `leave` properties
```
dylans: with an -> with

kitsonk:
This now requires dojo/shim#101 for the sets of features to actually be effective.

src/plugins/util/walk.ts
```diff
@@ -0,0 +1,60 @@
...
+				visit(element, node, enter, leave, key, idx);
+			});
+		}
+		else if (value && value.type) {
+			visit(value, node, enter, leave, key);
+		}
+	});
+
+	if (leave) {
+		leave(node, parent, prop, index);
```
rorticus: Probably not important, since it doesn't use `shouldSkip`, but it feels like `leave` should be called with the same context as `enter`? Just for consistency maybe.

kitsonk: Yeah, but we don't want people calling `this.skip()` thinking that it would do something.  So instead I have changed it to `null` and annotated the `this` in the types so that no one gets any bright ideas.

Issue #165, opened 2017-06-23
Integrate 3rd Party Libraries


kitsonk:
**Enhancement**

We need to articulate a workflow of how someone would integrate a 3rd party library into a Dojo 2 build.  Including libraries that are already in some sort of module format would be as simple as requiring them, but if there are other libraries or even built Dojo 1 layers that need to be made available at development time and at run-time, that workflow isn't entirely clear.

A specific use case would be how best to integrate ESRI's APIs which are distributed as a layer built with the Dojo 1 build tool.

PR #173, opened 2017-07-14
External dojo deps

maier49:
**Type:** feature

**Description:**

Partially addresses #165, by providing a mechanism for integrating Dojo 1 libraries, or any library that can be consumed with the Dojo 1 loader into a Dojo 2 application.


README.md
```diff
@@ -87,6 +87,37 @@ Ejecting `@dojo/cli-build-webpack` will produce a `config/build-webpack/webpack.
 node_modules/.bin/webpack --config=config/build-webpack/webpack.config.js
 ```
 
+### Interop with Dojo 1 libraries
+
+Dojo 1 libraries, whether built or not, can be included in a Dojo 2 application by configuring certain options in the project's `.dojorc` file.
+`.dojorc` is a JSON file that contains configuration for Dojo CLI tasks. Configuration for the `dojo build` task can be provided under the
```
dylans: Dojo CLI -> Dojo 2 CLI

README.md
```diff
@@ -87,6 +87,37 @@ Ejecting `@dojo/cli-build-webpack` will produce a `config/build-webpack/webpack.
...
+`build-webpack` property. Dojo 1 dependencies can be specified via a property called `externals` within the `build-webpack` config.
+`externals` is an object and the keys are paths, relative to `node_modules`, specifying the directories or files of the Dojo 1 dependencies that should
+be included. Each value can take three possible types:
```
dylans: Each value can take three... value isn't specific enough, also would suggest saying accept rather than take

README.md
```diff
@@ -87,6 +87,37 @@ Ejecting `@dojo/cli-build-webpack` will produce a `config/build-webpack/webpack.
...
+* `true`: Use this if none of the other options apply
```
dylans: None of what other options?

README.md
```diff
@@ -87,6 +87,37 @@ Ejecting `@dojo/cli-build-webpack` will produce a `config/build-webpack/webpack.
...
+* An array listing the package names that are in the dependency. This list is needed in order to tell the build process to leave the resolution of these
+files up to Dojo. If any module from a package is imported within application code but not specified in this list, it will fail at runtime.
```
dylans: files up to Dojo -> resources to Dojo 1

dylans: fail at runtime -> fail to load at runtime ?

README.md
```diff
@@ -87,6 +87,37 @@ Ejecting `@dojo/cli-build-webpack` will produce a `config/build-webpack/webpack.
...
+* An object with the following properties:
+ * `packages`: An array of package names, this array serves the same purpose as providing an array as the value.
+ * `hasLoader`: An optional flag to indicate that the specified layer includes a loader and a separate loader is not needed.
```
dylans: layer -> Dojo 1.x layer ?

README.md
```diff
@@ -87,6 +87,37 @@ Ejecting `@dojo/cli-build-webpack` will produce a `config/build-webpack/webpack.
...
+ * `main`: Can be provided if the external dependency is a folder. This is a path within the module to the file to be required. The folder or file
+ itself will be required if this is not specified. This is only relevant if `loadImmediately` or `hasLoader` is `true`.
```
dylans: is `true` -> is set to `true`

README.md
```diff
@@ -87,6 +87,37 @@ Ejecting `@dojo/cli-build-webpack` will produce a `config/build-webpack/webpack.
...
+ * `to`: An alternate location for a dependency to be copied to. The location is relative to the `externals`
```
dylans: An alternate location for a dependency to be copied to -> When copying a dependency, this specifies an alternate target location

README.md
```diff
@@ -87,6 +87,37 @@ Ejecting `@dojo/cli-build-webpack` will produce a `config/build-webpack/webpack.
...
+ folder. This can be useful if, for example, multiple built modules are being used that have shared dependencies, to ensure that only one copy
+ of each is included in the final build.
+ * `loadImmediately`: An optional flag that indicates this dependency should be loaded before the main application layer. The typical use case
+ for this is to load a layer file so that its contained modules will be available to the application. If a dependency has the loader,
+ `loadImmediately` is not needed as the loader script will be executed regardless.
+
+Types for any dependencies included in `externals` can be installed in `node_modules/@types` just like any other dependency.
+
+If the Dojo 1 library in question is a built layer, and has no external dependencies, then the `externals` config is all that's
```
dylans: then the `externals` config is all that's -> then `externals` is the only config

README.md
```diff
@@ -87,6 +87,37 @@ Ejecting `@dojo/cli-build-webpack` will produce a `config/build-webpack/webpack.
...
+needed to get a working build. However, in some cases not all dependencies are built into the layer file itself. This is often the case for i18n
+resources, with some plugins, if modules are intentionally excluded from a build, if multiple layers have shared dependencies, etc. In this case some
```
dylans: with some -> some

dylans: if modules are intentionally excluded from a build -> intentionally excluded modules within a build

dylans: if multiple layers have shared dependencies -> shared dependencies across layers

README.md
```diff
@@ -87,6 +87,37 @@ Ejecting `@dojo/cli-build-webpack` will produce a `config/build-webpack/webpack.
...
+additional configuration may be needed. To this end, the `externalConfig` property in the `.dojorc` configuration for `build-webpack` can be specified.
+This property will be passed as is to the loader to configure it. The [`dojoConfig` documentation](https://dojotoolkit.org/documentation/tutorials/1.10/dojo_config/)
+elaborates on the available configuration options.
```
dylans: available configuration -> available Dojo 1 configuration

src/plugins/ExternalDojoLoaderPlugin.ts
```diff
@@ -0,0 +1,158 @@
...
+const HtmlWebpackIncludeAssetsPlugin = require('html-webpack-include-assets-plugin');
+
+export type DojoDependencyDescriptor = {
+	/**
+	 * The path to the main file to be loaded. Can be provided if this dependency is a folder.
+	 * If not provided, `main.js` will be loaded by default.
+	 */
+	main?: string;
+	/**
+	 * Optional flag indicating whether this module includes a loader. If no module contains a loader an un-built Dojo
```
dylans: un-built Dojo -> un-built Dojo 1

src/plugins/ExternalDojoLoaderPlugin.ts
```diff
@@ -0,0 +1,158 @@
...
+	 * loader will be included in the build.
+	 */
+	hasLoader?: boolean;
+	/**
+	 * A List of packages defined by this dependency. This is used to populate the externals config. Any packages
+	 * that are defined by this dependency and imported in the application but not included in this list or will not
```
dylans: or will not ... or ?

src/plugins/ExternalDojoLoaderPlugin.ts
```diff
@@ -0,0 +1,158 @@
...
+	 * work. A DojoDependency that is an array of strings is equivalent to only defining this property.
```
dylans: this property -> this meaning hasLoader?

maier49:
@dylans I believe I've addressed all of your feedback.

matt-gadd:
@maier49 so I have a potentially stupid question, if i built my library on a way older version of dojo 1 (silly I know), and now want to include it I but have no loader in the layer. When we include a loader in the build, it will be from the version of `dojo` in cli-build package.json?

matt-gadd:
how would we build on this to support more external library types? could this be more generic in terms of it being an AMD externals configuration? (rather than dojo?). or even better decoupling ourselves completely kind of like intern 4 does in terms of it's loading mechanism and config?

maier49:
@matt-gadd To your question about the loader, if you don't include any layer with a loader, then yes you will get the version of `dojo` defined in the package.json. However there's nothing stopping you from building a separate layer that just includes the loader, or bringing along a specific version of `dojo` yourself and using the loader from that.

agubler:
I don't think we should be including Dojo as a dependency of cli-build, rather document that if the external needs a loader that it needs to be included specifically (if it's not built into the external itself).

maier49:
@agubler I've removed Dojo as a dependency

@matt-gadd I've reworked this to allow the registration of multiple loaders and left the loader config up to the developer. An example of providing a loader for Dojo 1 modules can be seen in the readme. Unfortunately the way that externals are loaded is pretty baked in, so this basically involved copying an existing plugin and selectively overriding functionality. See [here](https://github.com/webpack/webpack/issues/2280), and [here](https://github.com/webpack/webpack/issues/4171) for discussion in related issues.

src/plugins/ExternalLoaderPlugin.ts
```diff
@@ -0,0 +1,62 @@
...
+	apply(compiler: Compiler) {
+		compiler.apply(new CopyWebpackPlugin(this._externals.reduce(
+			(config, external) => typeof external === 'string' ? config : config.concat([ {
+				from: `node_modules/${external.from}`,
+				to: `externals/${external.to || external.from}`
+
+			} ]),
+			[] as { from: string, to: string }[]
+		)));
+		compiler.plugin(<any> 'this-compilation', (compilation: Compilation) => {
```
kitsonk: We should add this to the [WebPack typings](https://github.com/dojo/cli-build/blob/master/src/webpack.d.ts) that are part of the project so we don't have to cast to any here (or even type the `Compilation`).

src/plugins/ExternalLoaderPlugin.ts
```diff
@@ -0,0 +1,62 @@
+import Compiler = require('webpack/lib/Compiler');
+import Compilation = require('webpack/lib/Compilation');
+import Map from '@dojo/shim/Map';
```
kitsonk: Since our support for cli is Node.js 4+ and 4 had `Map` we shouldn't use shim here.  We should instead do like I have done in the [static build PR](https://github.com/dojo/cli-build/pull/172/files#diff-e5e546dd2eb0351f813d63d1b39dbc48) and add the collections lib to `tsconfig.json`

src/plugins/ExternalLoaderUmdTemplatePlugin.ts
```diff
@@ -0,0 +1,308 @@
+import Map from '@dojo/shim/Map';
```
kitsonk: As above, we shouldn't use Map shim with Node.js 4+

src/webpack.config.ts
```diff
@@ -3,7 +3,9 @@ import NormalModuleReplacementPlugin = require('webpack/lib/NormalModuleReplacem
 import * as path from 'path';
 import { existsSync, readFileSync } from 'fs';
 import { BuildArgs } from './main';
+import { ExternalDep } from './plugins/ExternalLoaderPlugin';
 import Set from '@dojo/shim/Set';
+import Map from '@dojo/shim/Map';
```
kitsonk: And again...

tsconfig.json
```diff
@@ -9,7 +9,7 @@
 		"outDir": "_build/",
 		"removeComments": false,
 		"sourceMap": true,
-		"target": "es6",
+		"target": "es5",
```
kitsonk: Was there a reason why we did this?  Just asking, as there likely is a good one.

matt-gadd:
Just trying to digest this. Is the reason for the copied plugin to support something other than UMD? Because the issues linked seem to be related to bespoke module formats, which I don't think we should bother supporting?

tsconfig.json
maier49: This was because we are now compiling a file that's run in the app `registerLoader`, and Uglify doesn't like es6.

maier49:
@matt-gadd The UMD plugin that I copied and overwrote is responsible for writing the UMD wrapper itself, and it automatically adds calls to `require` external modules.  That could be problematic in the case that, for example, we need to load a layer file before other external dependencies will be available, and couples us to using either global variables or a global `require` as the mechanism for loading external modules.

tsconfig.json
kitsonk: 👍

src/plugins/ExternalLoaderUmdTemplatePlugin.ts
```diff
@@ -167,7 +153,7 @@ export default class ExternalLoaderUmdTemplatePlugin {
 		const keys = Object.keys(customExternalLoaders);
 		const orderedExternalModules = keys.reduce((prev, next) => {
 			return prev.concat(customExternalLoaders[next]);
-		}, [] as any[]);
+		}, [] as Array<Loader | Module>);
```
kitsonk: Small nit, but we tend to use as `(Loader | Module)[]` for Array types.

README.md
```diff
@@ -87,6 +88,45 @@ Ejecting `@dojo/cli-build-webpack` will produce a `config/build-webpack/webpack.
...
+ | -------- | ---- | -------- | ----------- |
+ | `from` | `string` | false  | A path relative to `node_modules` that tells the loader where it can find this dependency in order to copy it into the built application |
+ | `to` | `string` | true | A path that replaces `from` as the location to copy this dependency to. Dependencies will be copied to `${externalsOutputPath}/${to}` or `${externalsOutputPath}/${from}` if `to` is not provided. |
+ | `name` | `string` | true | Indicates that this path, and any children of this path, should be loaded via the external loader |
+ | `inject` | `string | string[] | boolean` | true | This property indicates that this dependency defines, or includes, scripts or stylesheets that should be loaded on the page. If `inject` is set to `true`, then the file at the location specified by `to` or `from` will be loaded on the page. If this dependency is a folder, and then `inject` can be set to a string or array of strings to define one or more files to inject. Each path in `inject` should be relative to `${externalsOutputPath}/${to}` or `${externalsOutputPath}/${from}` depending on whether `to` was provided. |
```
agubler: Think that the pipes in the type column are messing with the formatting.

package.json
```diff
@@ -53,12 +53,14 @@
     "@dojo/has": "next",
     "@dojo/i18n": "next",
     "@dojo/shim": "next",
+    "auto-require-webpack-plugin": "^1.0.1",
```
agubler: let's pin this

src/plugins/CoreLoadPlugin.ts
```diff
@@ -58,7 +58,7 @@ function isContextual(module: NormalModule, issuers: string[]): boolean {
  * The updated path.
  */
 function stripPath(basePath: string, path: string): string {
-	let resolved = path.replace(basePath + '/', '').replace(/\..*$/, '');
+	let resolved = (basePath ? path.replace(basePath + '/', '') : path).replace(/\..*$/, '');
```
agubler: how come we needed this change?

src/plugins/ExternalLoaderPlugin.ts
```diff
@@ -0,0 +1,101 @@
...
+ */
+export type ExternalDep = string | ExternalDescriptor;
+
+export default class ExternalDojoLoaderPlugin {
+	private _externals: ExternalDep[];
+	private _outputPath: string;
+	private _pathPrefix: string;
+	private _hash: boolean;
+
+	constructor(options: {
```
agubler: could make a nice type for options and clean the formatting a little in the codez

agubler: Could in destructure in the params, like
```ts
constructor({ externals, outputPath: 'externals', pathPrefix, hash }: Options) {
     // stuff
}
```

src/plugins/ExternalLoaderPlugin.ts
```diff
@@ -0,0 +1,101 @@
...
+					return assets;
+				}
+
+				const base = to || from;
+
+				if (Array.isArray(inject)) {
+					return assets.concat(inject.map(path => prefixPath(`${base}/${path}`)));
+				}
+
+				const location = (typeof inject === 'string' && `${base}/${inject}`) || to || from;
```
agubler: Think this can be cleaner.

src/plugins/ExternalLoaderPlugin.ts
```diff
@@ -0,0 +1,101 @@
...
+	 */
+	to?: string;
+};
+
+/**
+ * Describes an external dependency
+ */
+export type ExternalDep = string | ExternalDescriptor;
+
+export default class ExternalDojoLoaderPlugin {
```
agubler: Can we drop Dojo from the naming?

src/plugins/CoreLoadPlugin.ts
maier49: When testing the generated ES6 code, the code coverage generation is actually dropping the default value `''` for basePath in the object destructuring. That was hiding a bug where if `basePath` was `''` we would remove the leading slash from a path. It was hidden because instead we were calling replace with `'undefined/'`. 

I discovered this when I switched to targeting ES5, although it's back to ES6 now so we would only see this if running tests without coverage.

README.md
```diff
@@ -87,6 +88,45 @@ Ejecting `@dojo/cli-build-webpack` will produce a `config/build-webpack/webpack.
 node_modules/.bin/webpack --config=config/build-webpack/webpack.config.js
 ```
 
+### Interop with external libraries
+
+External libraries that can not be loaded normally via webpack can be included in a Dojo 2 providing an implementation of `require` or `define`, and some
```
dylans: can not -> cannot

dylans: "can be included in a Dojo 2 providing" missing something here

README.md
```diff
@@ -87,6 +88,45 @@ Ejecting `@dojo/cli-build-webpack` will produce a `config/build-webpack/webpack.
...
+
+ `outputPath`: An optional property specifying an output path to which files should be copied.
+
+ `dependencies`: A required array that defines which modules should be loaded via the external loader, and what files should be included in the build. Each entry can be one of two types:
+* A string. In this case the entry simply indicates that this path, and any children of this path, should be loaded via the external loader
+* An object that provides additional configuration for dependencies that need to be copied into the built application. This object has the following properties:
+
+ | Property | Type | optional | Description |
+ | -------- | ---- | -------- | ----------- |
+ | `from` | `string` | false  | A path relative to `node_modules` that tells the loader where it can find this dependency in order to copy it into the built application |
```
dylans: "A path relative to `node_modules` that tells the loader where it can find this dependency in order to copy it into the built application" -> "A path relative to `node_modules` specifying the dependency location to copy into the built application" ?

README.md
```diff
@@ -87,6 +88,45 @@ Ejecting `@dojo/cli-build-webpack` will produce a `config/build-webpack/webpack.
...
+ | `to` | `string` | true | A path that replaces `from` as the location to copy this dependency to. Dependencies will be copied to `${externalsOutputPath}/${to}` or `${externalsOutputPath}/${from}` if `to` is not provided. |
```
dylans: "Dependencies will be copied to `${externalsOutputPath}/${to}` or `${externalsOutputPath}/${from}` if `to` is not provided." -> "By default, dependencies will be copied to `${externalsOutputPath}/${to}` or `${externalsOutputPath}/${from}` if `to` is not specified."

README.md
dylans: "and then `inject`" -> "then `inject`" ?

README.md
```diff
@@ -87,6 +88,45 @@ Ejecting `@dojo/cli-build-webpack` will produce a `config/build-webpack/webpack.
...
+ | `name` | `string` | true | Indicates that this path, and any children of this path, should be loaded via the external loader |
+ | `inject` | `string | string[] | boolean` | true | This property indicates that this dependency defines, or includes, scripts or stylesheets that should be loaded on the page. If `inject` is set to `true`, then the file at the location specified by `to` or `from` will be loaded on the page. If this dependency is a folder, and then `inject` can be set to a string or array of strings to define one or more files to inject. Each path in `inject` should be relative to `${externalsOutputPath}/${to}` or `${externalsOutputPath}/${from}` depending on whether `to` was provided. |
+
+ As an example the following config will inject `src/legacy/layer.js` into the application page, declare that modules `a`, `b`, and `c` are external
```
dylans: config -> configuration

README.md
```diff
@@ -87,6 +88,45 @@ Ejecting `@dojo/cli-build-webpack` will produce a `config/build-webpack/webpack.
...
+ and should be delegated to the external layer, and then copies over the folder `node_modules/legacy-dep`, from which several files are injected. All of these files will be copied into the `externals` folder, which could be overridden by specifying the `outputPath` property in the `externals` config.
```
dylans: "copies over the folder" -> "copy the folder"

dylans: config -> configuration

README.md
```diff
@@ -87,6 +88,45 @@ Ejecting `@dojo/cli-build-webpack` will produce a `config/build-webpack/webpack.
...
+        "a",
+        "b",
+        "c",
+        { "from": "src/legacy/layer.js", "inject": true },
+        { "from": "node_modules/legacy-dep", "inject": [ "modulA/layer.js", "moduleA/layer.css", "moduleB/layer.js" ] }
+    ]
+ }
+```
+
+Types for any dependencies included in `externals` can be installed in `node_modules/@types` just like any other dependency.
```
dylans: "just like" -> ", like"

maier49:
@dylans @agubler I've made updates to address your feedback.

README.md
```diff
@@ -88,6 +89,45 @@ Ejecting `@dojo/cli-build-webpack` will produce a `config/build-webpack/webpack.
...
+
+ `outputPath`: An optional property specifying an output path to which files should be copied.
+
+ `dependencies`: A required array that defines which modules should be loaded via the external loader, and what files should be included in the build. Each entry can be one of two types:
+* A string. In this case the entry simply indicates that this path, and any children of this path, should be loaded via the external loader
+* An object that provides additional configuration for dependencies that need to be copied into the built application. This object has the following properties:
+
+ | Property | Type | optional | Description |
+ | -------- | ---- | -------- | ----------- |
+ | `from` | `string` | `false`  | A path relative to `node_modules` specifying the depndency location to copy into the build application. |
```
dylans: depndency -> dependency

PR #177, opened 2017-07-19
Use different tsconfigs for building with and without tests

### LINKED ISSUE #113: --watch does not always recompile unit tests
**Bug**

When using --watch, the test source files are not always recompiled when they should.

Steps to reproduce:

1. Create an app: `dojo create --name test`
1. From the app directory, start the build: `dojo build --watch`.
1. Rename `src/App.ts` to `src/App1.ts`.
1. Notice the build watcher recompiles `src/main.ts` and `tests/unit/App.ts` and reports errors.
1. Edit `tests/unit/App.ts` and change the failing import statement to 
`import App from './../../src/App1';`
1. Notice no recompile occurs.
1. Edit `src/main.ts` and fix the import statement.
1. A recompile occurs but it still reports an error in `tests/unit/App.ts` even though the code had been updated.
1. Stop the build watch command and rerun it.
1. No compile errors.

Should the build watch command be compiling the test source files at all?  I am not going to answer that question.  But if there are cases when it is going to compile the files and report errors, then it needs to recompile them when I fix the errors.

maier49:

**Description:**
This updates the build so that it uses separate tsconfig files for `dojo build` and `dojo build -t`. `dojo build` would use a tsconfig file that only compiles sources files, and the tsconfig file for `dojo build -t` would also compile tests. Relies on dojo/cli-create-app#89 to create the separate config files in a new project.
Resolves #113

### DIFF
--- a/src/webpack.config.ts
+++ b/src/webpack.config.ts
@@ -250,7 +250,8 @@ function webpackConfig(args: Partial<BuildArgs>) {
 					{
 						loader: 'ts-loader',
 						options: {
-							instance: 'dojo'
+							instance: 'dojo',
+							configFileName: args.withTests ? 'tsconfig-tests.json' : 'tsconfig.json'
 						}
 					}
 				]},
@@ -272,7 +273,8 @@ function webpackConfig(args: Partial<BuildArgs>) {
 							{
 								loader: 'ts-loader',
 								options: {
-									instance: 'dojo'
+									instance: 'dojo',
+									configFileName: 'tsconfig-tests.json'
 								}
 							}
 						] }


matt-gadd:
@maier49 could we do this by just providing specific test related options to `ts-loader`? rather than putting a dependency on the end user having to have a non-standard extra tsconfig in their project (especially if they are not using create-app?)

maier49:
@matt-gadd It appears that ts-loader [ignores `include` or `exclude` options](https://github.com/TypeStrong/ts-loader/issues/544) and defers to tsconfig. I tested `include` locally to make sure. I understand the concern about requiring an extra config file though. Maybe there's another approach to this but I'm struggling to think of one at the moment.

matt-gadd:
Sorry I meant the ts-loader loader options (not the webpack loader options as in that issue). It looks like you can only provide the `compilerOptions` part of the `tsconfig` though. You can provide a separate `tsconfig` path as well though, so we could potentially write a temp file, but ideally it would be nice to just pass an object for the config, which we could perhaps explore as a PR to ts-loader?

agubler:
I think we should at least create a temporary tsconfig for tests "on the fly" if we can't get the preferred solution. I really don't want to have to have a none standard tsconfig file in order to use the build and test command.

maier49:
Closing in favor of generating a temp test config on the fly or deferring to config in ts-loader if they accept a PR.

PR #178, opened 2017-07-25
External css modules

### LINKED ISSUE #112: Can't load CSS modules from node modules

maier49:
**Type:** bug

**Description:**

This addresses a few issues that were preventing css modules in `node_modules` from being loaded.

* `css-module-dts-loader` wasn't able to find css module files in `node_modules`. This is fixed by using [`LoaderContext.resolve`](http://webpack.github.io/docs/loaders.html#resolve) to find the module.

* ` css-module-dts-loader` was overwriting the existing dts file with a version created from the `.m.css` file with generated classes. This is fixed by checking whether the css file is in the `src` directory and only generating a dts file if it is.

* The NormalModuleReplacementPlugin config in `webpack.config.ts`  wasn't properly resolving files in `node_modules`. The existing configuration resolves the request relative to the directory it was requested from when looking for an associated JavaScript file to load. This worked for widgets required from `@dojo/widgets` because the actual import of the css module is in the widget's JS file, and is a relative path. Checking for whether the requested path is relative, and resolving relative to `node_modules` if it is not fixes this.

* The NormalModuleReplacementPlugin config didn't work when requiring the same css file from multiple src files. The existing config assumes that two requests will be made for the same css file: One from the actual module that is importing it, and one from the `.m.css.js` file, which has a call to `require`  the css file added to it by `json-css-module-loader`. When importing the same module from two source files, the second request can come from the second source file and not the `.m.css.js` file. The requests generated by `json-css-module-loader` use absolute paths, so instead of using a flag to guess where the second request is coming from this just doesn't modify requests with absolute paths.

Resolves #112


bitpshr:
Looks great. I don't know the explicit requirements around minor things like how or if we should be rejecting if files can't be found, but the code is solid.

devpaul:
I'm having issues w/ the JLG project importing base styles from widgets & widget-core. If this fixes that I'd like to help test this and resolve it.

maier49:
This functionality has been ported over to dojo/webpack-contrib and these fixes are now part of a PR [there](https://github.com/dojo/webpack-contrib/pull/8) and a PR [here](https://github.com/dojo/cli-build-webpack/pull/230 ) to use that functionality.

PR #179, opened 2017-07-25
Adding check for port in use, issue #86

### LINKED ISSUE #86: Friendlier error message if you run multiple build-with-watch
If you run multiple `dojo build -w` then you'll get:

```
$ dojo build -w
events.js:160
      throw er; // Unhandled 'error' event
      ^

Error: listen EADDRINUSE 127.0.0.1:9999
    at Object.exports._errnoException (util.js:1022:11)
    at exports._exceptionWithHostPort (util.js:1045:20)
    at Server._listen2 (net.js:1262:14)
    at listen (net.js:1298:10)
    at doListening (net.js:1397:7)
    at _combinedTickCallback (internal/process/next_tick.js:77:11)
    at process._tickCallback (internal/process/next_tick.js:98:9)
```

Might be nice to capture this error and just say:

`Cannot start a build server because you already have one running.`

rorticus:
**Type:** feature

**Description:**

Adding an error message if you try to start the dev server and the port is in use. Note that this is not specifically checking if the watch command is already running, just that the specified port is available.

Resolves #86

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -100,7 +100,8 @@ async function isPortAvailable(port: number): Promise<boolean> {
 		server.once('error', function (err: any) {
 			if (err.code === 'EADDRINUSE') {
 				resolve(false);
-			} else {
+			}
+			else {
 				reject(new Error(`Unexpected error ${err.message}`));
 			}
 		});
--- a/src/main.ts
+++ b/src/main.ts
@@ -17,7 +17,7 @@ export interface BuildArgs {
 	messageBundles: string | string[];
 	supportedLocales: string | string[];
 	watch: boolean;
-	port: number;
+	port: string;
 	element: string;
 	elementPrefix: string;
 	withTests: boolean;
@@ -124,19 +124,42 @@ async function watch(config: webpack.Config, options: WebpackOptions, args: Buil
 	})(config.entry);
 
 	const compiler = webpack(config);
-	const port = args.port || 9999;
+	const portRange = String(args.port || '9999:9990');
+	let ports: number[] = [];
+	let serverPort: number | undefined;
 
-	const willSucceed = await isPortAvailable(port);
+	if (portRange.indexOf(':') >= 0) {
+		let [low, high] = portRange.split(':').map(p => parseInt(p, 10));
 
-	if (!willSucceed) {
-		return Promise.reject(new Error(`Cannot start a build server because port ${port} is in use. Do you already have a build server running?`));
+		if (high < low) {
+			[low, high] = [high, low];
+		}
+
+		for (let port = high; port >= low; port--) {
+			ports.push(port);
+		}
+	} else if (portRange.indexOf(',') >= 0) {
+		ports = portRange.split(',').map(p => parseInt(p, 10));
+	} else {
+		ports = [parseInt(portRange, 10)];
+	}
+
+	for (let i = 0; i < ports.length; i++) {
+		if (await isPortAvailable(ports[i])) {
+			serverPort = ports[i];
+			break;
+		}
+	}
+
+	if (!serverPort) {
+		return Promise.reject(new Error(`Cannot start a build server because the port is in use, tried ${ports.join(', ')}. Do you already have a build server running?`));
 	}
 
 	const server = new WebpackDevServer(compiler, options);
 
 	return new Promise<void>((resolve, reject) => {
-		server.listen(port, '127.0.0.1', (err: Error) => {
-			console.log(`Starting server on http://localhost:${port}`);
+		server.listen(serverPort, '127.0.0.1', (err: Error) => {
+			console.log(`Starting server on http://localhost:${serverPort}`);
 			if (err) {
 				reject(err);
 				return;
@@ -194,8 +217,8 @@ const command: Command<BuildArgs> = {
 
 		options('p', {
 			alias: 'port',
-			describe: 'port to serve on when using --watch',
-			type: 'number'
+			describe: 'port to serve on when using --watch. Can be a single port (9999), a range (9999:9990) or a list (9999,9997)',
+			type: 'string'
 		});
 
 		options('t', {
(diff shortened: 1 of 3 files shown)


matt-gadd:
@rorticus do you think we should try and select another port if the port is taken? rather than erroring? or beyond scope?

rorticus:
@matt-gadd feels like that's a little beyond scope. I like the idea a lot though.

agubler:
I don't really see what this adds over the existing default `EADDRINUSE` error message . However if we add in something smarter to select another port from a specified range then that would add value.

Erroring only if none of them are available

rorticus:
Well, it should be pretty easy to add. I can go ahead and upgrade this PR to do that if you want!

rorticus:
@agubler @matt-gadd Added the port searching stuff. By default it will use `9999:9990`, but you can also specify,

```
# search only on port 123
--port=123 

#search ports 9999, 9998, 9997, 9996
--port=9999:9996

#search ports 123, 456, 789
--port=123,456,789
```

src/main.ts
```diff
@@ -103,12 +125,42 @@ function watch(config: webpack.Config, options: WebpackOptions, args: BuildArgs)
...
+			[low, high] = [high, low];
+		}
+
+		for (let port = high; port >= low; port--) {
+			ports.push(port);
+		}
+	} else if (portRange.indexOf(',') >= 0) {
+		ports = portRange.split(',').map(p => parseInt(p, 10));
+	} else {
+		ports = [parseInt(portRange, 10)];
```
agubler: might look a bit nicer if you just push it on.

tests/unit/main.ts
```diff
@@ -140,10 +185,35 @@ describe('main', () => {
...
+	});
+
+	it('should run watch and reject on any listening error', () => {
+		const mockWebpackDevServer = mockModule.getMock('webpack-dev-server');
+		mockWebpackDevServer.listen = sandbox.stub().yields();
+		markPortAsInUse(true, 'SOMEERROR');
+		return moduleUnderTest.run(getMockConfiguration(), { watch: true }).then(
+			throwImmediately,
+			(e: Error) => {
+				assert.isTrue(e.message.indexOf('Unexpected error') >= 0);
```
agubler: presume we can actually just assert this real value (as we are mocking it?)

src/main.ts
```diff
@@ -103,12 +125,42 @@ function watch(config: webpack.Config, options: WebpackOptions, args: BuildArgs)
 	})(config.entry);
 
 	const compiler = webpack(config);
+	const portRange = String(args.port || '9999:9990');
+	let ports: number[] = [];
+	let serverPort: number | undefined;
+
+	if (portRange.indexOf(':') >= 0) {
+		let [low, high] = portRange.split(':').map(p => parseInt(p, 10));
```
agubler: should wrap argument in parenthesises, such as `((p) => parseInt(p, 10))`, also can do better at naming for `p`?

src/main.ts
```diff
@@ -103,12 +125,42 @@ function watch(config: webpack.Config, options: WebpackOptions, args: BuildArgs)
...
+
+		if (high < low) {
+			[low, high] = [high, low];
+		}
+
+		for (let port = high; port >= low; port--) {
+			ports.push(port);
+		}
+	} else if (portRange.indexOf(',') >= 0) {
```
agubler: `else` and `else if` need to be on a new line

src/main.ts
```diff
@@ -103,12 +125,42 @@ function watch(config: webpack.Config, options: WebpackOptions, args: BuildArgs)
...
 	const compiler = webpack(config);
+	const portRange = String(args.port || '9999:9990');
+	let ports: number[] = [];
+	let serverPort: number | undefined;
+
+	if (portRange.indexOf(':') >= 0) {
+		let [low, high] = portRange.split(':').map(p => parseInt(p, 10));
+
+		if (high < low) {
+			[low, high] = [high, low];
```
agubler: We don't cuddle in arrays or objects i.e. `[low, high]` or `{foo: 'bar'}` should be `[ low, high ]` or `{ foo: 'bar' }`

src/main.ts
```diff
@@ -103,12 +125,42 @@ function watch(config: webpack.Config, options: WebpackOptions, args: BuildArgs)
 	})(config.entry);
 
 	const compiler = webpack(config);
+	const portRange = String(args.port || '9999:9990');
+	let ports: number[] = [];
+	let serverPort: number | undefined;
+
+	if (portRange.indexOf(':') >= 0) {
```
agubler: Could probably add a const that describes this string (and the other delimiter)

src/main.ts
```diff
@@ -103,12 +125,42 @@ function watch(config: webpack.Config, options: WebpackOptions, args: BuildArgs)
 	})(config.entry);
 
 	const compiler = webpack(config);
+	const portRange = String(args.port || '9999:9990');
```
agubler: Same for the default port range, lets add a const for it.

rorticus:
@agubler actually i'm pretty sure my styles *aren't* set up right 😄  I'll be setting rules going forwards so hopefully the silly errors go away.

Also, I made all the changes you requested!

src/main.ts
```diff
@@ -84,7 +94,28 @@ function mergeConfigArgs(...sources: BuildArgs[]): BuildArgs {
...
 
-function watch(config: webpack.Config, options: WebpackOptions, args: BuildArgs): Promise<void> {
+async function isPortAvailable(port: number): Promise<boolean> {
+	const server = net.createServer();
+
+	return new Promise<boolean>((resolve, reject) => {
+		server.once('error', function (err: any) {
+			if (err.code === 'EADDRINUSE') {
+				resolve(false);
+			} else {
```
agubler: missed this one.

rorticus:
Fixed 😄

agubler:
@rorticus can you add a little bit to the readme?

rorticus:
@agubler Updated the README

PR #186, opened 2017-07-28
Fix using command on Windows

### LINKED ISSUE #174: Fix tests on Windows
Several unit tests currently fail on Windows due to path issues.

rorticus:
**Description:**

@maier49 's PR here, https://github.com/dojo/cli-build-webpack/pull/176 but rebased onto master.

Resolves #174

### DIFF
--- a/src/plugins/CoreLoadPlugin.ts
+++ b/src/plugins/CoreLoadPlugin.ts
@@ -140,7 +140,7 @@ export default class DojoLoadPlugin {
 	constructor(options: DojoLoadPluginOptions = {}) {
 		const { basePath = '', chunkNames, detectLazyLoads, ignoredModules, mapAppModules = false } = options;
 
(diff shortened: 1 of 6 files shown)


src/plugins/CoreLoadPlugin.ts
```diff
@@ -140,7 +140,7 @@ export default class DojoLoadPlugin {
 	constructor(options: DojoLoadPluginOptions = {}) {
 		const { basePath = '', chunkNames, detectLazyLoads, ignoredModules, mapAppModules = false } = options;
 
-		this._basePath = basePath;
+		this._basePath = basePath && basePath.replace(/\\/g, '/').replace(/^[cC]:/, '');
```
agubler: why do you need `basePath && ...` as it is defaulted to `''` above?

Issue #207, opened 2017-08-30
Compilation and linting errors do not cause building to fail


mwistrand:
**Bug**

If there are linting or TS compiler errors in the source code, running `$ dojo build` does not fail to compile, allowing type errors and formatting issues to more easily slip through. It would be ideal if both errors resulted in a failed build, at least when not in watch mode.

(dojo/cli-test-intern#31 is a similar issue)

PR #208, opened 2017-09-01
Compliation errors cause the build to fail now, dojo/cli-test-itern#31

rorticus:
**Type:** bug

**Description:**

Errors during the webpack build will now return a rejected promise. With this, `dojo test` will fail when you have build errors.

Resolves https://github.com/dojo/cli-test-intern/issues/31

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -36,6 +36,7 @@ export interface BuildArgs {
 	bundles: Bundles;
 	externals: { outputPath?: string; dependencies: ExternalDep[] };
 	features: string | string[];
+	force: boolean;
 }
 
 interface ConfigFactory {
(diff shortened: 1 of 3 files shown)


src/main.ts
```diff
@@ -196,6 +197,14 @@ function compile(config: webpack.Config, options: WebpackOptions): Promise<void>
 				}
 
 				console.log(stats.toString(options.stats));
+
+				if (stats.compilation && stats.compilation.errors && stats.compilation.errors.length > 0) {
+					reject({
+						exitCode: 1,
+						message: ''
```
kitsonk: surely we can provide something more informative?

rorticus: Well.. I tried something more informative but it just looked silly. At this point the webpack build has already printed out all the errors so the console is full of the _real_ error details.  I had could try something like `The build failed with errors` or something, but like I said, it seemed obvious from the red text above it... @kitsonk thoughts?

kitsonk: Hmm... slightly raises the question of if we have a `--force` or not... which then this message could indicate that there are compilation errors reported, use `--force` to ignore.

rorticus: I like that. I will add support for `--force`! Can I print out `use the --force to continue anyways`? 😛

kitsonk: These aren't the 🤖 you are looking for.

mwistrand:
This also addresses at least part of #207, correct?

rorticus:
Good catch @mwistrand , yes!

PR #209, opened 2017-09-05
Adding test bundle

### LINKED ISSUE #127: Different chunks for tests and src and dependencies?
**Enhancement**

It would be good if we built tests, src and dependencies as different chunks when doing a dev build.  The reason for this is that when instrumenting with Istanbul via Intern, basically Istanbul ends up instrumenting the whole bundle and collecting data.  This causes a significant amount of data to be collected which is then subsequently dropped.  Outside of the performance impact in instrumenting and collecting the data, large sets of coverage data can cause issues in certain browsers when posting the data back from a remote session.  I suspect any decent sized application would blow up IE11, Safari, and iOS, browsers we have had difficulty collecting coverage data from when they are too large.  End users would be clue-less to what the problem is other than staring at their remote session apparently hanging.

rorticus:
**Type:** feature

**Description:**

Adding separate bundles for code being tested and "support" code. The code being tested is run through istanbul for coverage and is ready for intern 4!

This is used by https://github.com/dojo/cli-test-intern/pull/61

Resolves #127


rorticus:
Removed the separate bundles and am just using the same bundle we've always had, but using istanbul to manually instrument only the `src/**/*.ts` files in the bundle.

src/webpack.config.ts
```diff
@@ -240,17 +258,28 @@ function webpackConfig(args: Partial<BuildArgs>) {
...
+							return module.context && module.context.indexOf('src/') !== -1;
+						}
+					})];
 			}),
 			...includeWhen(includesExternals, () => [
 				new ExternalLoaderPlugin({
 					dependencies: externalDependencies,
 					outputPath: args.externals && args.externals.outputPath,
-					pathPrefix: args.withTests ? '../_build/src' : ''
+					pathPrefix: args.withTests ? `../_build/src` : ''
```
matt-gadd: i don't think we needed to change this line

src/webpack.config.ts
```diff
@@ -240,17 +258,28 @@ function webpackConfig(args: Partial<BuildArgs>) {
 					]),
 					new HtmlWebpackPlugin ({
 						inject: true,
-						chunks: [ '../_build/src/main' ],
+						chunks: [ 'src', `../_build/src/main` ],
```
matt-gadd: can this be a normal string now too?

src/webpack.config.ts
```diff
@@ -143,6 +152,16 @@ function webpackConfig(args: Partial<BuildArgs>) {
...
 			new webpack.BannerPlugin(readFileSync(require.resolve(`${packagePath}/banner.md`), 'utf8')),
+			/**
+			 * We're using the banner plugin here to fix a bug with the webpack jsonp function. When the function is used
+			 * in the test bundle, the variable is not scoped at all, so it has, in our case, a function scope in node.
+			 * This little hack makes sure the function is defined by grabbing it from the window scope (jsdom).
+			 */
+			new webpack.BannerPlugin(<any> {
+				banner: `var ${jsonpFunctionName} = ${jsonpFunctionName} || window["${jsonpFunctionName}"];`,
+				raw: true,
+				test: /tests\/unit\/all\.*/
```
matt-gadd: what are we doing about functionals?

src/webpack.config.ts
```diff
@@ -240,17 +258,28 @@ function webpackConfig(args: Partial<BuildArgs>) {
...
+						chunks: [ 'src', `../_build/src/main` ],
 						template: 'src/index.html',
-						filename: '../_build/src/index.html'
-					})
-				];
+						filename: `../_build/src/index.html`
+					}),
+					new webpack.optimize.CommonsChunkPlugin({
+						name: 'src',
+						filename: `../_build/src/src.js`,
```
matt-gadd: can be just a string

src/webpack.config.ts
```diff
@@ -240,17 +258,28 @@ function webpackConfig(args: Partial<BuildArgs>) {
...
+						chunks: [`../_build/src/main`, '../_build/tests/unit/all'],
```
matt-gadd: can be just a string

src/webpack.config.ts
```diff
@@ -240,17 +258,28 @@ function webpackConfig(args: Partial<BuildArgs>) {
...
+						minChunks: (module: any) => {
+							if (module.resource && !(/^.*\.(ts)$/).test(module.resource)) {
```
matt-gadd: what does this do again?

src/webpack.config.ts
rorticus: Welp. The functional tests shouldn't be loading the source module, so they shouldn't be an issue. If we want to support that, we can load that `src/src.js` module for functional tests too,as well as apply this little hack. I don't know that there is any specific down side to supporting it but I left it out since its a sort of an anti-pattern I think?

matt-gadd: yeah that seems reasonable to me 👍

src/webpack.config.ts
rorticus: So this is the part that actually extracts the source bundle. So basically `src/*/*.ts` will be taken out of `src/main` and `tests/unit/all` and placed in this `src.js` bundle.  The source is not extracted from the functional tests here because we're _assuming_ that the functional tests are only loading things like `.css` files.

rorticus:
In regards to the istanbul loader, I talked to @jason0x43 and it sounds like this will be eventually moved to intern. Added to https://github.com/theintern/intern/issues/769 .

rorticus:
OK, issue raised in Webpack and tracked by https://github.com/dojo/cli-build-webpack/issues/211

src/loaders/istanbul-loader/index.ts
```diff
@@ -0,0 +1,3 @@
+import loader from './loader';
```
kitsonk: showing as uncovered

src/webpack.config.ts
```diff
@@ -82,10 +86,12 @@ function webpackConfig(args: Partial<BuildArgs>) {
 		});
 	}
 
+	const jsonpFunctionName = getJsonpFunction(packageJson && packageJson.name);
```
kitsonk: uncovered line

src/webpack.config.ts
```diff
@@ -240,11 +258,22 @@ function webpackConfig(args: Partial<BuildArgs>) {
...
 						filename: '../_build/src/index.html'
-					})
-				];
+					}),
+					new webpack.optimize.CommonsChunkPlugin({
+						name: 'src',
+						filename: '../_build/src/src.js',
+						chunks: ['../_build/src/main', '../_build/tests/unit/all'],
+						minChunks: (module: any) => {
+							if (module.resource && !(/^.*\.(ts)$/).test(module.resource)) {
```
kitsonk: more uncovered

jason0x43:
A beta of the [istanbul-loader](https://github.com/theintern/istanbul-loader) is now out as `@theintern/istanbul-loader`. Install the 'next' version if you want to try it out ('latest' is also a beta, just the first one).

PR #219, opened 2017-10-06
Fix assumptions about path separator

### LINKED ISSUE #218: Various build errors on Windows
**Bug** 

There are a number of issues when building on Windows due to the fact that `cli-build-webpack` does not thoroughly account for the correct path separator:

* `umd-compat-loader` does not replace the path with bundle name.
* `InjectModulesPlugin` is unable to locale modules injected at build time.
* `CoreLoadPlugin` does not map modules correctly.
* `I18nPlugin` does not include locale-specific bundles in the build.

Package Version: 2.0.0-beta3.1

mwistrand:
**Type:** bug

**Description:**

Changes have been verified with `todo-mvc-kitchensink` on both Windows and Mac. Updates operations that read and manipulate file paths to correctly account for the OS's file separator. Fixes a number of bugs on Windows:

* `umd-compat-loader` now uses the bundle's name.
* `CoreLoadPlugin` maps module IDs correctly.
* `InjectModulesPlugin` is able to load the specified modules.
* Supported locale bundles are properly included in the build.

Resolves #218

### DIFF
--- a/src/plugins/CoreLoadPlugin.ts
+++ b/src/plugins/CoreLoadPlugin.ts
@@ -57,7 +57,7 @@ function isContextual(module: NormalModule, issuers: string[]): boolean {
  * @return The normalized file path.
  */
 function normalizeFilePath(path: string): string {
-	return path.replace(/\\/g, '/').replace(/^[cC]:/, '');
+	return path.replace(/\\/g, '/').replace(/^[a-z]:/i, '');
 }
 
 /**
--- a/src/plugins/CoreLoadPlugin.ts
+++ b/src/plugins/CoreLoadPlugin.ts
@@ -1,5 +1,5 @@
 import * as path from 'path';
-import { getBasePath, resolveMid } from './util/main';
+import { createFilePathRegExp, getBasePath, resolveMid } from './util/main';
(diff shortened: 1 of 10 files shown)


src/plugins/CoreLoadPlugin.ts
```diff
@@ -38,15 +44,26 @@ const jsMidPattern = /\.(t|j)sx?$/;
...
+/**
+ * @private
+ * Strips the drive prefix from a file path and normalizes the path separator to '/'.
+ *
+ * @param path The path on which to operate.
+ *
+ * @return The normalized file path.
+ */
+function normalizeFilePath(path: string): string {
+	return path.replace(/\\/g, '/').replace(/^[cC]:/, '');
```
tomdye: couldn't the users machine be using any letter for the windows installation? not just `c` ?.
Aren't there filepath resolving util libraries available for this?

mwistrand: Good catch; the drive can be any letter. I'd copied that from elsewhere in the module. To answer your question, `path` methods do not convert from `\` to `/`.

PR #234, opened 2017-10-27
Use new ts-loader and compatible version of webpack-contrib

### LINKED ISSUE #227: Prevent build from failing for TS errors in files outside of webpack bundles
**Enhancement** 

Currently TS compilation errors in any file in the paths configured in tsconfig will fail the build, even if they are not included in the bundles. This manifests itself as annoying and sometimes unintuitive failures when test files fail to compile, or when `m.css` files are depended on by files that aren't included in the build.

ts-loader now supports transformers which may provide an avenue for resolving this.

maier49:
**Type:** bug

**Description:**
This updates the version of ts-loader used, and uses the new `onlyCompileBundledFiles` option. As a result, errors in test files, or other files that are not included in the resulting bundle, will not fail `dojo build`. `dojo build -w` still does not rebuild when tests are built, but `dojo build -w --with-tests` does, and since test files no longer break `dojo build -w` that's ok.

The other changes are just fixing some weird formatting issues in `webpack.config.ts`, and updating to the next version of webpack-contrib (not released yet) that will fix an incompatibility between the css-module-dts-loader and the new version of ts-loader.

Resolves #227, #113


maier49:
@matt-gadd Is this still something we'd like to land here or should it just wait for the new build command to be completed?

matt-gadd:
@maier49 lets land this okay.

PR #240, opened 2017-11-09
Upgrade to TypeScript 2.6 and turn on strict

bryanforbes:
**Type:** enhancement

**Description:**

Relates to dojo/meta#210

### DIFF
--- a/src/plugins/CoreLoadPlugin.ts
+++ b/src/plugins/CoreLoadPlugin.ts
@@ -330,9 +330,9 @@ export default class DojoLoadPlugin {
 				}
 			});
 
-			compilation.moduleTemplate.plugin('module', (source, module: NormalModule) => {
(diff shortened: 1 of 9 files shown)


src/plugins/CoreLoadPlugin.ts
```diff
@@ -330,9 +330,9 @@ export default class DojoLoadPlugin {
 				}
 			});
 
-			compilation.moduleTemplate.plugin('module', (source, module: NormalModule) => {
+			compilation.moduleTemplate.plugin('module', (source, module) => {
```
agubler: How come removing this type and then casting?

bryanforbes: Strict function typing doesn't allow typing `module` as `NormalModule` in the function declaration

PR #247, opened 2017-11-27
Fix: parse `@dojo/i18n/cldr/load` requires from sequence expressions.

### LINKED ISSUE #246: CLDR paths are not parsed when building with tests
**Bug**

When building with the `--withTests` option, the CLDR paths are not parsed from the source and therefore no CLDR data are injected into the build.

Package Version: 0.2.1

**Code**

```
import loadCldrData from '@dojo/i18n/cldr/load';

loadCldrData([ 'path/to/cldr/data.json' ]);
```

**Expected behavior:**

The build should parse and load `path/to/cldr/data.json` into the build.

**Actual behavior:**

The build does not parse the URL and injects only an empty object.

mwistrand:
**Type:** bug

**Description:**

`@dojo/i18n/cldr/load.default` can be required from within a sequence expression, which appears to be the case when building an application with the `--withTests` flag. This PR adds a check for such calls, and has been verified against the [`todo-mvc-kitchensink`](https://github.com/dojo/examples) tests.

Resolves #246

### DIFF
--- a/src/plugins/util/i18n.ts
+++ b/src/plugins/util/i18n.ts
@@ -37,7 +37,7 @@ function getCallExpressionData(expression?: any): { callee?: Identifier, args?: 
 		return {};
 	}
 
-	const { callee, 'arguments': args } = expression;
+	const { callee, arguments: args } = expression;
 	return { callee, args };
 }
 
--- a/src/plugins/util/i18n.ts
+++ b/src/plugins/util/i18n.ts
@@ -1,8 +1,45 @@
 import Map from '@dojo/shim/Map';
-import { CallExpression, Identifier, Program, VariableDeclaration, VariableDeclarator } from 'estree';
+import { CallExpression, Expression, Identifier, Program, SequenceExpression, SpreadElement, VariableDeclaration, VariableDeclarator } from 'estree';
 import * as path from 'path';
 import { getBasePath, isRelative, mergeUnique } from './main';
 import { extractArrayValues, getNextItem, isShadowing } from './parser';
+
+/**
+ * @private
+ * Determines whether the provided expression is a CallExpression.
+ */
+function isCallExpression(expression: any): expression is CallExpression {
+	return expression.type === 'CallExpression';
+}
+
+/**
+ * @private
+ * Determines whether the provided expression is a SequenceExpression.
+ */
+function isSequenceExpression(expression: any): expression is SequenceExpression {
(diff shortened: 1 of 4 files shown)


src/plugins/util/i18n.ts
```diff
@@ -1,9 +1,46 @@
...
+ * @param expression An AST expression node.
+ * @return An object containing the callee and arguments if the expression is a call expression.
+ */
+function getCallExpressionData(expression?: any): { callee?: Identifier, args?: Array<Expression | SpreadElement> } {
+	if (isSequenceExpression(expression)) {
+		const callExpressions = expression.expressions.filter(expression => isCallExpression(expression)) as CallExpression[];
+		expression = callExpressions[0];
+	}
+
+	if (!expression) {
```
agubler: There don't seem to be any unit tests to cover this branch

mwistrand: Thanks, @agubler. I've added a test that covers this branch.

src/plugins/util/i18n.ts
```diff
@@ -1,9 +1,46 @@
...
+
+/**
+ * @private
+ * Resolves the callee and arguments for the provided expression.
+ *
+ * @param expression An AST expression node.
+ * @return An object containing the callee and arguments if the expression is a call expression.
+ */
+function getCallExpressionData(expression?: any): { callee?: Identifier, args?: Array<Expression | SpreadElement> } {
+	if (isSequenceExpression(expression)) {
```
agubler: What happens if `expression` is `undefined` or `null`, wouldn't you get `Cannot read property 'type' of undefined` when you call `isSequenceExpression`?

mwistrand: Dang it, you're right. I've pushed another update.

src/plugins/util/i18n.ts
```diff
@@ -1,9 +1,46 @@
...
+		const callExpressions = expression.expressions.filter(expression => isCallExpression(expression)) as CallExpression[];
+		expression = callExpressions[0];
+	}
+
+	if (!expression) {
+		return {};
+	}
+
+	const { callee, 'arguments': args } = expression;
```
agubler: do we need the quotes? would just straight `const { callee, arguments: args } = expression;` work?

mwistrand: That would work, it was just an arbitrary personal preference on my part.

PR #250, opened 2017-11-30
Optimize all js files

### LINKED ISSUE #244: --feature flag no longer working
**Bug** 

At some point, it seems the `--feature` and `-f` no longer do anything.

Package Version: 0.2.1

**Code**

```
$ dojo version

The currently installed groups are:

build (@dojo/cli-build-webpack) 0.2.1
create (@dojo/cli-create-app) 0.1.0

You are currently running @dojo/cli 0.2.0

$ dojo build
$ ls -lh dist/src/main.js
-rw-r--r--  1 kitsonk  staff   135K 23 Nov 13:25 dist/src/main.js
$ dojo build --features chrome
$ ls -lh dist/src/main.js
-rw-r--r--  1 kitsonk  staff   135K 23 Nov 13:26 dist/src/main.js
```

**Expected behavior:**

The build with `--features chrome` to be smaller than the build not using the flag.

**Actual behavior:**

The builds are the same size.

maier49:
**Type:** bug

**Description:**
Previously, the static optimizer plugin has access to all JS files by virtue of the fact that it was accessing the compilation. When I moved it to a loader, I only added it to the `src/**.ts` files. This updates it to properly optimize any JS files in the build.

Resolves #244

### DIFF
--- a/src/webpack.config.ts
+++ b/src/webpack.config.ts
@@ -345,9 +345,6 @@ function webpackConfig(args: Partial<BuildArgs>) {
 				{ test: /@dojo\/.*\.js$/, enforce: 'pre', loader: 'source-map-loader-cli', options: { includeModulePaths: true } },
 				{ test: /src[\\\/].*\.ts?$/, enforce: 'pre', loader: '@dojo/webpack-contrib/css-module-dts-loader?type=ts&instanceName=0_dojo' },
 				{ test: /src[\\\/].*\.m\.css?$/, enforce: 'pre', loader: '@dojo/webpack-contrib/css-module-dts-loader?type=css' },
-				{ test: /\.js(x)?$/, loader: '@dojo/webpack-contrib/static-build-loader', options: {
-					features: args.features
-				}},
 				{ test: /src[\\\/].*\.ts(x)?$/, use: [
 					{
 						loader: '@dojo/webpack-contrib/static-build-loader',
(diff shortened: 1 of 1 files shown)


src/webpack.config.ts
```diff
@@ -345,6 +345,9 @@ function webpackConfig(args: Partial<BuildArgs>) {
 				{ test: /@dojo\/.*\.js$/, enforce: 'pre', loader: 'source-map-loader-cli', options: { includeModulePaths: true } },
 				{ test: /src[\\\/].*\.ts?$/, enforce: 'pre', loader: '@dojo/webpack-contrib/css-module-dts-loader?type=ts&instanceName=0_dojo' },
 				{ test: /src[\\\/].*\.m\.css?$/, enforce: 'pre', loader: '@dojo/webpack-contrib/css-module-dts-loader?type=css' },
+				{ test: /\.js(x)?$/, loader: '@dojo/webpack-contrib/static-build-loader', options: {
```
kitsonk: I don't believe we will ever have `.jsx` files emitted... The TypeScript compiler will only ever emit `.js` files for us.  If for some reason we did, we wouldn't want to touch them I suspect.

maier49: Ok. I added the loader to the existing loader config for all `js` files and removed this separate entry entirely.

PR #262, opened 2018-01-04
Changed dojo cli flag to a string

### LINKED ISSUE #261: process.env.DOJO_CLI should be a string
**Enhancement** 

According to typings for node process flags, they should be `string | undefined` and not a boolean. This will need to be changed in `cli-build-webpack` along with the supporting change in `@dojo/cli`.

tomdye:
**Type:** bug / feature

* [ ] All code matches the [style guide](https://github.com/dojo/meta/blob/master/STYLE.md)

**Description:**
Updated typings in CLI require the DOJO_CLI env flag to be a string.

Resolves #261
Requires: https://github.com/dojo/cli/issues/188


tomdye:
This is likely no-longer required as it only effects version `0.6.0+` of `@dojo/cli` at which point this package will be deprecated.

PR #266, opened 2018-01-31
Use interfaces from `@dojo/cli` instead of `@dojo/interfaces`

tomdye:
**Type:** bug / feature

**Description:**

Uses interfaces from cli instead of interfaces package

Addresses https://github.com/dojo/meta/issues/223


agubler:
`@dojo/cli-build-webpack` has been deprecated

PR #268, opened 2018-05-01
pin @types/node version to ~9.6.5

nicknisi:
This is tracked in dojo/meta#247.

### DIFF
--- a/tests/unit/main.ts
+++ b/tests/unit/main.ts
@@ -29,7 +29,7 @@ describe('main', () => {
 	}
 
 	beforeEach(() => {
-		process.env.DOJO_CLI = true;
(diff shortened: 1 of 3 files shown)


tests/unit/main.ts
```diff
@@ -29,7 +29,7 @@ describe('main', () => {
 	}
 
 	beforeEach(() => {
-		process.env.DOJO_CLI = true;
+		process.env.DOJO_CLI = 'true';
```
maier49: Is this to satisfy the updated types?

maier49 (approved):

tests/unit/main.ts
nicknisi: Yep

